PR #8, opened 2015-03-13
Fixed a bug where Gum would crash on start.

AnsisMalins:
The previous behavior was that if the project file is not found, Gum would just go on trying to use the uninitialized gps variable and crash with NullReferenceException. Also, checking if file exists and then doing something is susceptible to race conditions. So I replaced all that with a try-catch block.


bjorn:
Looks fine to me.

PR #21, opened 2015-03-30
Removed the SaveProject call in LoadProject

### LINKED ISSUE #18: Project is saved when it is opened
I guess it's a measure to upgrade files to any new format changes, but this happens even when there are no changes to save. This behavior wreaks havoc with Perforce. Gum should only save when it makes sense.

bjorn:
It's in my opinion quite a bad idea to automatically save the whole project when it gets loaded. Two examples where this goes wrong:

* Version control systems like Perforce require files to be checked out before being written to. In effect, this behavior causes all files in the project to be checked out just by opening Gum.

* Sometimes I've managed to start Gum without our plugin loaded. This would then inadvertedly cause the Standard elements to get written to disk, including the set of properties that I'm suppressing in the   plugin.

In general, I think it's just unexpected behavior. It's also unnecessary because whatever changes are being made to the project during load, you can expect Gum to make those same changes again next time it loads the project.

Finally, if you really want a project to be saved in its entirety, then there is already a menu item available for this.

Closes #18.

### DIFF
--- a/Gum/DataTypes/ElementSaveExtensionMethods.cs
+++ b/Gum/DataTypes/ElementSaveExtensionMethods.cs
@@ -45,21 +45,16 @@ namespace Gum.DataTypes
 
     public static class ElementSaveExtensionMethods
     {
-        public static bool Initialize(this ElementSave elementSave, StateSave defaultState)
+        public static void Initialize(this ElementSave elementSave, StateSave defaultState)
         {
-            bool wasModified = false;
-
-            if (AddAndModifyVariablesAccordingToDefault(elementSave, defaultState))
-            {
-                wasModified = true;
-            }
+            AddAndModifyVariablesAccordingToDefault(elementSave, defaultState);
 
             foreach (StateSave state in elementSave.AllStates)
             {
                 state.ParentContainer = elementSave;
                 state.Initialize();
 
-                FixStateVariableTypes(elementSave, state, ref wasModified);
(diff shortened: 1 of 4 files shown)


vchelaru:
If I remember correctly Gum does not always save the project upon load, but instead simply checks if the project is missing variables that it should have (such as if new variables have been added to the standard objects).

If Gum does not automatically save the project upon load when it finds missing values, then the user should be notified of this upon opening. Otherwise users who are not familiar with the inner-workings of Gum will not be aware of these changes and the requirement to manually save the project.

I think that we should modify this code to do two things:

1. It should selectively save the changed files rather than save the entire project. Then the only modified files (and only files checked out by Perforce) will be the files that are necessary in regards to the underlying modification.
2. If we remove the auto-save on load, then we should have some way to notify the user something like "Stanard elements X, Y, and Z have changed, would you like to save these now?"

bjorn:
Hmm, I don't agree that it makes sense to notify the user about these changes, because I don't think it makes sense to save at all, not manually either. This is mainly because next time you load the project, the same changes will happen anyway, so the need to save is simply not there.

As far as I can tell the need to save the standard components is only there because you could use this to change the default values of their properties. And this functionality behaves the same whether updated files are saved on load or not.

vchelaru:
I believe this has been fixed in the current Gum. Sorry for not grabbing this pull request.

bjorn:
No problem!

PR #22, opened 2015-04-01
Added (After)ElementSave plugin event

AnsisMalins:
I realized I need one more event for the Perforce plugin to work correctly. It's named ElementSave to match the ProjectSave event.

### DIFF
--- a/Gum/Gui/Plugins/MainWindowPlugin.cs
+++ b/Gum/Gui/Plugins/MainWindowPlugin.cs
@@ -35,7 +35,7 @@ namespace Gum.Gui.Plugins
         public override void StartUp()
         {
             ProjectLoad += new Action<DataTypes.GumProjectSave>(OnProjectLoad);
-            ProjectSave += new Action<DataTypes.GumProjectSave>(OnProjectLoad);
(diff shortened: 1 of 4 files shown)


vchelaru:
Could we change the name to "AfterElementSave" as the element has already been saved? I don't know if there is a BeforeElementSave but I could see plugins needing before and after logic on saving.

bjorn:
I think the point is, if "AfterElementSave" is better then "ProjectSave" should be renamed to "AfterProjectSave" as well. I think in either case, it is not a reason not to merge this.

AnsisMalins:
Implemented Bjorn's suggestion.

bjorn:
My suggestion was to either rename neither or both. Now you renamed one of them. :-)

AnsisMalins:
What? It's now BeforeProjctSave, AfterProjectSave, BeforeElementSave, and AfterElementSave.

bjorn:
@AnsisMalins Ah, I was looking at `ProjectLoad` and wondered why it wasn't `AfterProjectLoad` now.

bjorn:
Personally I prefer `ProjectSaved` rather than `AfterProjectSave`, btw.

AnsisMalins:
Naming convention flamewar go!

PR #26, opened 2015-04-09
Made some parts of DeleteObjectPlugin public

AnsisMalins:
I've made DeleteXmlComboBox and GetFileNameForObject public so other plugins can access this information. It is necessary for my Perforce plugin so it can remove the XML file from Perforce correctly.


vchelaru:
I think this is okay, but before I merge I just wanted to ask - what properties on the combo box do you need access to? Does it make sense to simply expose the properties off of the combo box that are needed rather than the entire combo box?

If a lot of the properties are needed, or if you envision other plugins needing access to the entire combo box let me know and I'll merge.

AnsisMalins:
You are right. I do not foresee anyone needing more than the Checked property, so I changed the code accordingly.

PR #75, opened 2022-03-28
Update to HTTPS URLs

patridge:
I've updated the links on the [Gum intro page on flatredball.com](https://flatredball.com/documentation/tools/gum/gum-tutorials/tutorials-gum-introduction-and-setup/) and wanted to do the same here as well to avoid browser warnings on users' browsers.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,9 +2,9 @@
 
 ## Downloading Gum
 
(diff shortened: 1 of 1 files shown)


vchelaru:
Thanks!!

PR #92, opened 2023-10-07
Moving an object now takes RestrictToUnitValues into account

Arcnor:
When moving (and resizing) an object, the "RestrictToUnitValues" is only honored after the fact, which is a bit jarring in my opinion (you move it to some place, but after you let go of your mouse it slightly moves back to some invisible grid).

I've added code to fix this issue for movement, although I need some advice on how to proceed:
1. Code is very verbose, not sure if there is a leaner way of doing this. Moving once (instead of moving + adjusting) is obviously possible, but I used the same code from `StandardWireframeEditor::SnapSelectedToUnitValues()` for simplicity, as I'm not even sure of how to fetch the actual "X/Y" value given the levels of indirection present :).
2. Given the resizing code might also move the object to compensate, that one is even more difficult. It might be easier to be able to:
    a. Get existing coordinate value
    b. Adjust to integer value
    c. Set the new value instead of incrementing relatively.
...but this might not be possible at all, hence I'm asking for advice or hep on how to fix that part as well.

Thanks!

### DIFF
--- a/Gum/Wireframe/EditingManager.cs
+++ b/Gum/Wireframe/EditingManager.cs
@@ -136,40 +136,70 @@ namespace Gum.Wireframe
                 if (SelectedState.Self.SelectedInstances.Count() == 0 && 
                     (SelectedState.Self.SelectedComponent != null || SelectedState.Self.SelectedStandardElement != null))
                 {
+                    var selfSelectedElement = SelectedState.Self.SelectedElement;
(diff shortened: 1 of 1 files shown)


vchelaru:
Closing this PR as this was fixed with a different approach

91803e35e2120809ad2f476ed5ab79db44adb24a

PR #93, opened 2023-10-08
Replace color picker with Pixi.ColorPicker

Arcnor:
This is probably a "personal preference" issue, so feel free to close if you don't think it's useful.

Personally, I think the current color picker is very hard to use, both because of the unordered list of colors as well as the lackluster "advanced" mode, which doesn't even have a way of picking colors using the HSV/HSL spaces. Also you need 2 clicks to access it, which is not good design (again, IMHO).

I've replaced it with a color picker I found that seems to be an improvement over it, albeit probably still not perfect: https://github.com/PixiEditor/ColorPicker/

Gum_NRTsvem42Z

### DIFF
--- a/Gum/Controls/ColorDisplay.xaml.cs
+++ b/Gum/Controls/ColorDisplay.xaml.cs
@@ -12,6 +12,7 @@ using System.Windows.Media;
 using System.Windows.Media.Imaging;
 using System.Windows.Navigation;
 using System.Windows.Shapes;
+using ColorPicker.Models;
 using WpfDataUi;
 using WpfDataUi.DataTypes;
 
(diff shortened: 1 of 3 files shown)


Gum/Controls/ColorDisplay.xaml.cs
```diff
@@ -154,12 +155,17 @@ private void HandlePropertyChange(object sender, PropertyChangedEventArgs e)
...
         }
 
-        private void HandleColorChange(object sender, RoutedPropertyChangedEventArgs<Color?> e)
-        {
+        private void HandleColorChange(object sender, RoutedEventArgs e) {
+            if (!(e is ColorRoutedEventArgs colorArgs)) return;
+
+            var prevColor = (Microsoft.Xna.Framework.Color)InstanceMember.Value;
+
+            var colorPack = (uint)(colorArgs.Color.R | (colorArgs.Color.G << 8) | (colorArgs.Color.B << 16) | (colorArgs.Color.A << 24));
```
Arcnor, line 163: This check is necessary because `Pixi.ColorPicker` seems to fire the `ColorChanged` event multiple times per change (around 8 if I'm not mistaken), which makes the picker painful to use. There might be a better way to solve this, but I'm no WPF developer so I'd appreciate any feedback there.

vchelaru: I'm not too familiar with this picker myself, but if this restricts the spamming of events, then I think it's a good solution. I'll approve it and see how it works.

PR #102, opened 2023-10-08
Add child/parent Instance to instances (as opposed to only Elements)

Arcnor:
Context menu entries for adding child and parent to `Instances`.
Gum_1MPnHORnkE

Something missing is expanding the added node after the fact, I'm not sure how to retrieve the `TreeNode` from the `InstanceSave`, so the new node defaults to be closed.

### DIFF
--- a/Gum/Managers/NameVerifier.cs
+++ b/Gum/Managers/NameVerifier.cs
@@ -244,25 +244,25 @@ namespace Gum.Managers
 
             if (objectToIgnore != elementSave && name == elementSave.Name)
             {
-                whyNotValid = "The element is named " + elementSave.Name;
+                whyNotValid = $"The element is named '{elementSave.Name}'";
             }
 
             var instance = elementSave.Instances.FirstOrDefault(item => item != objectToIgnore && item.Name == name);
             if (instance != null)
             {
-                whyNotValid = "There is already an instance named " + instance.ToString();
+                whyNotValid = $"There is already an instance named '{instance.Name}'";
             }
 
             var state = elementSave.States.FirstOrDefault(item => item != objectToIgnore && item.Name == name);
             if (state != null)
             {
-                whyNotValid = "There is already a state named " + state.ToString();
+                whyNotValid = $"There is already a state named '{state.Name}'";
             }
             
             var variable = elementSave.AllStates.SelectMany(item => item.Variables).FirstOrDefault(item => item != objectToIgnore && item.ExposedAsName == name);
             if (variable != null)
             {
-                whyNotValid = "There is already a variable named " + variable.ToString();
+                whyNotValid = $"There is already a variable named '{variable.Name}'";
             }
 
             //element = ObjectFinder.Self.GumProjectSave.StandardElements.FirstOrDefault(item=>item != objectToIgnore && item.Name == name)
--- a/Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.RightClick.cs
+++ b/Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.RightClick.cs
@@ -15,6 +15,7 @@ using ToolsUtilities;
 using Gum.Wireframe;
 using Gum.Logic;
 using Gum.DataTypes.Behaviors;
+using Gum.PropertyGridHelpers;
 
 namespace Gum.Managers
 {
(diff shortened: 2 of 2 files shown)


Gum/Managers/NameVerifier.cs
```diff
@@ -244,25 +244,25 @@ private void IsNameAlreadyUsed(string name, object objectToIgnore, ElementSave e
 
             if (objectToIgnore != elementSave && name == elementSave.Name)
             {
-                whyNotValid = "The element is named " + elementSave.Name;
+                whyNotValid = $"The element is named '{elementSave.Name}'";
```
Arcnor, line 247: This (and the rest of the checks) was also modified because the message was sometimes a bit confusing, but maybe there is a good reason why it was like this, so I'd leave that to your judgement.

Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.RightClick.cs
```diff
@@ -672,28 +683,81 @@ private void HandleAddLinkedComponentClick(object sender, EventArgs e)
...
+        {
+            // From DragDropManager:
+            // "Since the Parent property can only be set in the default state, we will
+            // set the Parent variable on that instead of the SelectedState.Self.SelectedStateSave"
+            var stateToAssignOn = targetElement.DefaultState;
+
+            var variableName = newInstance.Name + ".Parent";
+            var existingInstanceVar = existingInstance.Name + ".Parent";
+            var oldValue = stateToAssignOn.GetValue(variableName) as string;        // This will always be empty anyway...
+            var oldParentValue = stateToAssignOn.GetValue(existingInstanceVar) as string;
```
vchelaru, line 742: This method and the one below should respect the default child container. To get a child container, you can call

ObjectFinder.Self.GetDefaultChildName(instanceSave)

If the value returned is not null or empty, then the parent should append "." + returnFromGetDefaultChildName. This will respect the default child name when adding here.

Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.RightClick.cs
```diff
@@ -672,28 +683,81 @@ private void HandleAddLinkedComponentClick(object sender, EventArgs e)
...
+            stateToAssignOn.SetValue(variableName, oldParentValue, "string");
+            stateToAssignOn.SetValue(existingInstanceVar, newInstance.Name, "string");
 
+            SetVariableLogic.Self.PropertyValueChanged("Parent", oldValue, newInstance);
+            SetVariableLogic.Self.PropertyValueChanged("Parent", oldParentValue, existingInstance);
+        }
+
+        private static void SetInstanceParent(ElementSave targetElement, InstanceSave child, InstanceSave parent)
+        {
+            // From DragDropManager:
```
vchelaru, line 753: See above for discussion about GetDefaultChildName

Gum/Managers/NameVerifier.cs
vchelaru: This validation check should read something to the effect of:

"The element containing this is named {elementSave.Name}, so this cannot be named the same name."

In C#, a member in a class cannot be named the same as the class itself. Since Gum is often used to codegen, naming something within an Element the same name as the element itself will cause compile errors.  Gum could ignore this, but we enforce this rule in Gum for the sake of errors downstream.

Please feel free to change this error message to be clearer, using the phrasing above if you feel it makes sense.

Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.RightClick.cs
Arcnor: Hey, will try to do as you say, although I'm a bit confused as to what you mean exactly here :). I think I copied this from some other part of the code that changes the children parent, and also in this case the child will always have no parent because we just created it, right? Not sure if that affects what you said anyway :D.

Gum/Managers/NameVerifier.cs
Arcnor: Yep, maybe I should have commented on the next lines, as this line didn't change much, but after this one, I've replaced `instance.ToString()` with `instance.Name` for example.

Not sure if 1. that can cause any issues and 2. is something you want, mainly because now instead of "Container Thing of Test" you'll only get "Thing" (because at least for the errors I got, having the whole phrase was a bit strange).

vchelaru: If that reads better, that's fine by me.

PR #167, opened 2024-06-25
Gum.Kni

nkast:


ToolsUtilities/FileManager.cs
```diff
@@ -21,16 +21,29 @@ public static partial class FileManager
...
+                try
+                {
+                    return Path.GetDirectoryName(AppContext.BaseDirectory)
+                            .Replace('/', Path.DirectorySeparatorChar)
+                            .Replace('\\', Path.DirectorySeparatorChar) + Path.DirectorySeparatorChar;
+                }
+                catch 
+                {
+                    // WASM ?
+                    return "./";
```
nkast, line 42: AppContext.BaseDirectory throws in WASM, and I changed ExeLocation to "./".
I am not sure if that is the correct fix here.

nkast:
DesktopGL & WindowsDX seems to work.
BlazorGL will run without exceptions, but tries to load content from `http://localhost:5259/Content/Content/BearTexture.png`
and sprites/textures will not show. The correct address of the file is 'http://localhost:5259/Content/BearTexture.png'

MonoGameGum/KniGum.csproj
```diff
@@ -0,0 +1,100 @@
...
+    <PackageId>Gum.KNI</PackageId>
+    <Version>2024.6.20.2</Version>
+
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+
+
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DefineConstants>$(DefineConstants);MONOGAME;KNI;USE_GUMCOMMON</DefineConstants>
```
nkast, line 17: Here I define both MONOGAME and KNI. Perhaps it's best to define only KNI and expand all 
`#if MONOGAME` 
to
`#if MONOGAME || KNI`

ToolsUtilities/FileManager.cs
vchelaru: I think that's okay, unless we can get a proper #if

MonoGameGum/KniGum.csproj
vchelaru: Ya I think that's probably best, even though it's more work.

ToolsUtilities/FileManager.cs
nkast: Actually it was Path.GetDirectoryName() that was failing.
Now I skip it when the value is "/", and let the rest of the method to do, whatever it does.
This also fixed the issue we had with /Content/Content/ path!

MonoGameGum/KniGum.csproj
nkast: Done.

nkast:
KniGumInCode

Here is the latest KniGumInCode on the web.
The issue with /Content/Content/ disappeared once we have the proper fix for ExeLocation.

nkast:
KniGumFromFile

Here is the latest KniGumFromFile on WindowsDX.

GumDataTypes/GumProjectSave.cs
```diff
@@ -422,21 +422,24 @@ private void PopulateElementSavesFromReferences(string projectRootDirectory, Lin
                 }
             }
 
-            foreach(var reference in BehaviorReferences)
+            if  (BehaviorReferences != null)
```
nkast, line 425: I tried rooting the Gum assemblies from the main project, and setting IsTrimmable on the libraries.
That seems to be XmlSerializer behavior.

Issue #144, opened 2024-01-29
Search should also do text searches


vchelaru:
Currently it is possible to search for items by name:

![image](https://github.com/vchelaru/Gum/assets/729631/b32b2262-2959-44d6-9258-02388258205c)

It's common to look for UI based on the text on that UI. For example, searching for a button that says "Save". 

This may be slower because it requires searching all variables to see if the Save text is assigned. Therefore, we should have some kind of additional option so that it doesn't happen on all searches. Perhaps an "Include text" or something?

Alternatively, we could have a Find menu item... Need to think about this...

vchelaru:
Closed by https://github.com/vchelaru/Gum/pull/177

PR #177, opened 2024-08-21
Text search and search item hierarchy

akaadream:
This PR got a ref to #144

The user can now search deeply inside the projet using a checkbox.
The PR also add a quality of life for the found elements with the path of the element.
Clicking on a found element directly focus it.

### DIFF
--- a/Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.cs
+++ b/Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.cs
@@ -1699,7 +1699,7 @@ namespace Gum.Managers
 
             deepSearchCheckBox = new CheckBox();
             deepSearchCheckBox.Checked = false;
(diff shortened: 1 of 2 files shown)


Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewManager.cs
```diff
@@ -1562,6 +1612,18 @@ private void ReactToFilterTextChanged()
             //}
         }
 
+        private void AddToFlatList(object element, string customName = "")
+        {
+            if (element == null)
+            {
+                throw new Exception("The element should not be null");
```
vchelaru, line 1619: Can you `throw ArgumentNullException(nameof(element));` instead?

akaadream:
I updated the PR with the ArgumentNullException and the name of the element as the exception message.

Issue #130, opened 2023-12-28
[Feature Request] Allow making a system wide component from an existing screen element


KallDrexx:
Sometimes you set up elements on one screen, and realize you want to repeat those elements on other screens.  You can manually copy/paste from one screen to the other, but it can also make sense to make a system wide component out of that element, so that it's centrally managed for all screens it's relevant on.

This is important because it's not trivial to see the values you set for an existing element when trying to recreate it for a new component.

vchelaru:
This concept is similar to the concept in Visual Studio where you can select a block of code and extract method.

Phase 1:

1. Right-click on an instance in a screen or a component
2. Select the option "Create Component"
3. A popup appears similar to the add component popup which asks the user for the name of the component. For now, we can make this just go into the root folder and we can add folder support later.
4. When the user clicks OK, a new component is created with the desired name from the popup.
5. The component should have its base type be set to the type of the selected instance from step 1. This should be initially tested with Container as the base type since that's the most common type, and then test it with more complex types later.
6. Each child of the selected instance from step 1 should be copied in to the newly-created component
7. Each variable from the default state for any of the copied instances should be copied over to the new component.
8. Copy variables from the instance onto the newly-created component. 

Phase 2:

1. Add a checkbox that the user can select "Replace {InstanceName} and all children with an instance of the new component
2. If checked, and if the user selects OK, delete all children of the instance that was selected and the instance itself
3. Add a new instance with the same name that is of the type of the newly created component. All variables on the instance itself should remain.

Notes:

kaltinril:
This appears to be implemented, no?  What is missing?

vchelaru:
It is missing creating a component that includes all children of what is selected. I think it only currently considers the top level object

vchelaru:
Sounds good. Can we make phase 1 and phase 2 as separate PRs?

Pituivan:
Sure, since they're technically two different features (I already made a prototype and I had to create another menu action because when you analyze it, Create Component and this feature request are two completely separate things that simply have some inner logic in common).

I called this feature "Extract Component" in my prototype.

PR #179, opened 2024-08-25
Create a component base on element instances #130

akaadream:
_This PR is referencing the issue #130_

This PR is created for testing the feature.
There is still few issues which prevent the new component to be rendered correctly on the preview (a lot of variables are missing in the .gumc).


Gum/Commands/EditCommands.cs
```diff
@@ -496,6 +496,7 @@ public void CreateComponent()
                     if (string.IsNullOrEmpty(whyNotValid))
                     {
                         ComponentSave componentSave = new ComponentSave();
+                        componentSave.BaseType = "Container";
```
akaadream, line 499: This line fix the issue of the component save file to not include a lot of variables (so the preview could not display the instances)

akaadream:
There is still few fix to do (I'll update this post if needed):
* If we create a component of an instance which got children, children are not included inside the new component

Gum/Commands/EditCommands.cs
```diff
@@ -465,6 +466,81 @@ public void DuplicateSelectedElement()
 
         }
 
+        public void CreateComponent()
```
vchelaru, line 469: This method doesn't always create the component - it depends on error checking and if the user clicks OK. I would suggest naming this "ShowCreateComponentFromInstancesDialog" or something to that effect to indicate that the dialog is being shown but that the component may not actually be created.

Gum/Commands/EditCommands.cs
```diff
@@ -465,6 +466,81 @@ public void DuplicateSelectedElement()
...
+                            componentSave.States.Add(state.Clone());
+                        }
+
+                        StandardElementsManagerGumTool.Self.FixCustomTypeConverters(componentSave);
+                        ProjectCommands.Self.AddComponent(componentSave);
+                        
+                    }
+                    else
+                    {
+                        MessageBox.Show($"Invalid name for new component: {whyNotValid}");
```
vchelaru, line 538: Should we re-open the dialog to let the user correct the problem? Otherwise if they have an error, they have to do the right-click command again.

Gum/Commands/EditCommands.cs
```diff
@@ -465,6 +466,81 @@ public void DuplicateSelectedElement()
...
+            }
+            else if (instances is List<InstanceSave>)
+            {
+                TextInputWindow tiwcw = new TextInputWindow();
+                tiwcw.Message = "Enter new Component name:";
+
+                FilePath filePath = element.Name;
+                var nameWithoutPath = filePath.FileNameNoPath;
+
+                tiwcw.Result = nameWithoutPath;
```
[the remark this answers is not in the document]
akaadream: The new component now have "Component" at the end of its name.

The scan of the project will require a bit more works.

PR #182, opened 2024-09-05
Test gum gui namespace

akaadream:


akaadream:
Just a reminder for me to know the tasks for the rest of the PR (actually, each task could be a unique PR):
* Create a WPF Form for the component creating based on instances. (This form should know when the name of the component is not available and disable the validation button in the case of already used name).
* Add an option to replace the old instances with the new created component.
* Currently, the user have to select all the instances before creating the component even for instances which are children of another instance. We should add the children of an instances automatically without having to select all the instances.

vchelaru:
Merging now as we've discussed that the remaining features could be their own PR.

PR #198, opened 2024-09-29
make combobox theme friendly

vicdotexe:


vchelaru:
There's a problem with setting non-default values which I'll explain. I'd like to understand how this is supposed to work better so I can help maintain it, but I'm not sure how to solve this bug the "right" way.

Currently, if you select an object in Gum that has a default state, then set it to non-default, the text remains green. You can deselect and re-select the object, in which case the combo box does show the correct color, but if you keep the object selected and toggle between default and non-default, the text remains green. See this gif:



I believe the code responsible for setting the foreground to black is here, which is supposed to happen when the foreground property is cleared:

```cs
private void SyncForegroundWithState()
{
    if (InstanceMember.IsDefault)
    {
        ComboBox.Foreground = Brushes.Green;
    }
    else
    {
        ClearValue(ComboBox.ForegroundProperty);
    }
}
```

I have verified that when the property is explicitly set the property is cleared, but the text remains green. 

I do not understand how ClearValue is supposed to work - are properties a stack, and setting the Foreground to green pushes on the stack, then ClearValue pops off the stack restoring the previous value? That would seem to be implied by the code I'm seeing but for some reason it isn't working.

For comparison I did look at CheckBoxDisplay and I see that it does something fairly similar:

```cs
private void SetForeground(Brush brush)
{
    if (brush == Brushes.Black)
    {
        CheckBox.ClearValue(Control.ForegroundProperty);
    }
    else
    {
        CheckBox.Foreground = brush;
    }
}
```

It is doing it based on whether the value is black or not, but the effect should be the same. That seems to work fine; yet, it does not for ComboBoxDisplay and I'm not sure why.

vicdotexe:
Without being at the computer, I'd imagine where I broke it is I removed the block at line 274 and didn't replace it with the new method.. 

As far as clear value goes, foreground is a dependency property, when you clear the value you're saying "this property is not being set by the control" and it will walk up and grab the  value next in line, in this case the next value is coming from the implicit style...

vchelaru:
I did add this:
image
It didn't fix the problem. Also, the SyncForegroundWithState is being called, I verified with a breakpoint.

vchelaru:
Got it, we weren't setting ClearValue on the ComboBox itself.

PR #211, opened 2024-10-14
Move the creation logic into the null check

### LINKED ISSUE #193: Project Properties should not be able to be opened more than 1 time

kaltinril:
Fixes #193

I just moved the items that create a new instance of the control into the null check


kaltinril:
This turns out it would actually cause issues because if someone clicked Close on the properties window, they would never be able to bring it back up, I've canceled my PR

PR #268, opened 2024-11-24
1642 - hint-text wrapping

vicdotexe:
fixes https://github.com/vchelaru/FlatRedBall/issues/1642

### DIFF
--- a/WpfDataUi/Controls/NullableBoolDisplay.xaml
+++ b/WpfDataUi/Controls/NullableBoolDisplay.xaml
@@ -20,7 +20,8 @@
                     TextWrapping="Wrap"
                     HorizontalAlignment="Left"
                     VerticalAlignment="Top"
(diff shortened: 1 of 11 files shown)


WpfDataUi/Controls/NullableBoolDisplay.xaml
```diff
@@ -20,7 +20,8 @@
                     TextWrapping="Wrap"
                     HorizontalAlignment="Left"
                     VerticalAlignment="Top"
-                    Width="{Binding ElementName=TopRowGrid, Path=ActualWidth, Mode=OneWay}"
+                    MaxWidth="{Binding ActualWidth, ElementName=TopRowGrid}"
```
vicdotexe, line 23: @vchelaru 
This control doesn't appear to actually have an element named "TopRowGrid", I just left the element name as-is incase it was somehow working in a way you knew of that I couldn't see... mostly because I was too lazy to figure out how to throw a nullable bool in the variables. You may want/need to circle back to this and modify it to use the "GroupBox" element name.

PR #363, opened 2024-12-17
Add a check to see if the listview has polygon points, and if so, res…

### LINKED ISSUE #361: BUG: (Low) Polygon - Should not be able to delete all the points if deleting from the Points List
Currently if you click on a point and press delete, you can delete all but that last 3 points (4 technically, but the last is a repeat of the first).

However, the Points ListView lets you delete as many points as you want.

The ListView should behave the same way the Editor does and limit your deletions to not delete all the points.

**Steps to reproduce:**
1. Drag a polygon to a component or screen
2. Click in the Points ListView
3. Press delete
4. Click and delete again and again until no points remain
5. Now you can no longer use the EDITOR to add points as there are no line segments.

**Fix:**
Prevent the user from deleting the last 3 points.

This is **low priority**, a user can just UNDO or ADD a fresh polygon.

https://github.com/user-attachments/assets/70625304-0a79-49d4-abd5-653a671bc1ed

kaltinril:
…trict deletion to leave 4 points

Fixes #361

Video shows me adding a polygon, adding some points, then trying to delete all the points, (Delete key press not visible obviously)

https://github.com/user-attachments/assets/fc448773-a7a7-41a2-ad46-253bb0168161

### DIFF
--- a/WpfDataUi/Controls/ListBoxDisplay.xaml.cs
+++ b/WpfDataUi/Controls/ListBoxDisplay.xaml.cs
@@ -193,10 +193,15 @@ namespace WpfDataUi.Controls
                 if(selectedItem > -1)
                 {
                     var listToRemoveFrom = ListBox.ItemsSource as IList;
-
+                    
                     if(ListBox.SelectedIndex < listToRemoveFrom.Count)
                     {
-                        listToRemoveFrom.RemoveAt(ListBox.SelectedIndex);
+                        // Don't let a user delete the last 3 points in a Polygon (4th point is to close the shape)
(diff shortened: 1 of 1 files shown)


kaltinril:
My code change felt a bit hacky, if you would prefer a different/better way please do

vchelaru:
Unfortunately we can't do this because the List display is a general-purpose UI element which could be used in any situation. There aren't cases which use Lists of Vector2's yet, but there could be in the future. We need to have this check somewhere higher up in the Gum code rather than the list display window code.

kaltinril:
Sounds good, canceling this PR

Issue #405, opened 2024-12-24
BUG: File Watch path doesn't update when you create a new project (It should)


kaltinril:
Currently if you create a new project and save it, the File Watch path remains at whatever the previous project was.  This is an issue and should NOT happen.  It SHOULD update to the new Projects path.

Steps to reproduce:
1. Open Gum
2. If a project has not yet been created on this computer, create and save a new project
3. If a project was already open, continue.
4. Click View > Show File Watch
5. Click New Project
6. Notice that the File Watch path remains at whatever the OTHER Project folder path was
7. Click Save project and save it
8. Notice again, file watch path did not update
9. Create some components, instances, etc
10. Notice no changes to file watch path.

https://github.com/user-attachments/assets/802fbee7-0320-426e-afc2-6e805f7e1305

vchelaru:
I believe this is fixed now

PR #408, opened 2024-12-24
Add enabled for FileWatchManager and FileWatchLogic.  Add call to HandleProjectLoaded inside SaveProject

kaltinril:
Attempt at #405


kaltinril:
@vchelaru I couldn't get it to work correctly.  

1. Couldn't figure out how to force the File Paths watching to get emptied out on New Project
2. I was able to get a Save Project to call the functionality that gets the root project path
3. However... when I added a screen and components, it didn't have standard or screen or component folder it was watching, so I didn't do something correctly in this PR.

So there must be other locations.

vchelaru:
I think the changes you made were good. One of the biggest problems with the code (as it was before the PR) is that it was not using the plugin pattern that new code is using, so like you said the code was spread across lots of files and the flow was hard to follow.

I moved most of the code reacting to project load/unloads/renames into the plugin, and also moved some files related to file watch into the plugin folder as well. 

I'll be happy to answer any questions or help resolve confusion.

Issue #413, opened 2024-12-27
UX: (Minor) Behavior - Import window


kaltinril:
- [ ] 1. Popup window is missing a header/title
- [ ] 2. The window is a bit confusing, and, the text for the prompt is small and hard to read at 1440p.
- [ ] 3. The text says "The list above shows all files in the project folder" which is a lie, it shows nothing.
- [ ] 4. Suggestion: Grey out OK until file is selected (Can skip if you want)
- [ ] 5. What is the point of the single line Textbox at the top?
- [ ] 6. If the popup for selecting a file outside the project is going to remain, change No to cancel.  No doesn't explain what will happen, cancel implies better.
- [ ] 7. DOC: Add information about this import feature

![image](https://github.com/user-attachments/assets/25a3ee01-97cd-4790-8e90-790c2de6f067)

![image](https://github.com/user-attachments/assets/9997676b-ac3e-4612-a1f5-ad1affb678aa)

kaltinril:
Current state of the window
![Image](https://github.com/user-attachments/assets/d2b0556d-ccc1-42f6-8871-434c5d7237b3)

kaltinril:
I've marked item 6 in my list as done, because I honestly don't know what I was talking about.

kaltinril:
Ok, I think the wording on the bottom should change to

"The list above shows all <components> in the project folder that have not yet been imported."

I also think the search textbox at the top needs to have the word "search" before it, because I wasn't sure what the heck it was for

PR #436, opened 2024-12-28
Add the title for the Import Behavior window

kaltinril:
Related to #413


kaltinril:
Actually, going to cancel this because I didn't realize this control was reused, fixing it in the massive title add instead.

Issue #468, opened 2025-01-04
UI Scaling Fallout compilation (Compilation ticket)


kaltinril:
This is going to be a compilation ticket, when an item is going to be worked, an ISSUE should be created so spawn off the work, and linked here.  When the work for that sub-ticket is complete, click the checkmark next to the item.

**Items:**
- [ ] 1. (high) [functional] Animations - Interpolation values - Text grows and is cut off, pushing Textbox and drop downs off the screen.  This section is not manually expandable by the user.  So that may also want to get addressed.
![image](https://github.com/user-attachments/assets/7bcf3342-d573-41c5-b6cb-86faa25b3149)
- [ ] 2. (low) [visual] Animations - Speed Adjustment control (- 100% +) text becomes hard to read about 7 zoom ins (probably close to 300%)
![image](https://github.com/user-attachments/assets/85d78429-7251-403d-b00d-d82870ba830d)
- [ ] 3. (low) [visual] Animations - Keyframes text - Does not adjust with UI scale
![image](https://github.com/user-attachments/assets/97b03c48-8102-4e15-841a-6dc03a5d1ffc)
- [ ] 4. (mid) [functional] Animations - Interpolation values - First combobox doesn't grow
![image](https://github.com/user-attachments/assets/3cda5e26-7b1c-44e5-988b-fcadab6b2818)
- [ ] 5. (low) [visual] Treeview - Search variables checkbox - This has a small width, instead of being the width of the parent, which causes the text to wrap and get cut off.  Suggest changing this control into a WPF control and setting it to full width.
![image](https://github.com/user-attachments/assets/111dfefd-459a-4b79-becc-7e221f4a9c65)
- [ ] 6. (mid) [visual] State/Category - Folder icons - They do not grow or shrink.  Look into doing the same thing that was done for the primary TreeView
![image](https://github.com/user-attachments/assets/e9271dc3-3828-4703-b7ed-65ae900087d1)
- [ ] 7. (mid) [visual] Variables - First column width - Currently it is hard coded and so text wraps ugly.  Make this fixed width adjust based on the UI zoom/scale percent perhaps?
![image](https://github.com/user-attachments/assets/609c03ee-4f2f-4feb-b733-1c6dbc7c0490)
- [ ] 8. (low) [visual] TreeView Project Search texbox cancel X button outside panel
![image](https://github.com/user-attachments/assets/d3af2296-1e3c-4f0e-a5e7-d827661969b3)
[...]

PR #469, opened 2025-01-04
Fix search checkbox width. Fix X button width.  Fix file selection buttons size scaling.

kaltinril:
This fixes a few items, but more are needed for #468

items 5, 8, 11, and 13 are fixed or greatly enhanced from that issue.


kaltinril:
Image showing 3 of the 4 fixed
image


Image showing the 4th item greatly enhanced (Still a little alignment issue with the degrees, but at least you can read it all now).
image

PR #503, opened 2025-01-16
commented out problematic line, Vic needs to change some more stuff

vchelaru:


vchelaru:
I need to check whether children are null before climbing up to find the item with the layer.

Issue #545, opened 2025-01-30
Variable References - Math - Multiplication


kaltinril:
I did

Red = Blue * 5

and gum crashed

https://github.com/user-attachments/assets/e2cf03a9-c4f1-4d06-9f1f-c64a4de80e36

vchelaru:
There are 2 problems here:

1. You are referencing `Blue` which in this case means `Comp.Blue` (I have a card about whether this behavior is desirable
2. Gum should not crash, and should provide a more informative error. I've fixed this with 3350a797c5471f0b30582126b7cf43eefa6497f7

Issue #544, opened 2025-01-30
Variable References - Setting X to Red crashes gum


kaltinril:
I set X equal to Red and it crashed gum

X = Red

https://github.com/user-attachments/assets/892447d5-6c3a-4122-b740-aeb2599def01

kaltinril:
It appears the colors are somehow being evaluated as empty/null and being set to 

Red = 

when EvaluatedSyntax.FromSyntaxNode is called

vchelaru:
This has been fixed. It's happening because Red does not exist on the component (you are assigning an instance's X to the parent component's Red)

I have another issue about whether this behavior should change, but strictly speaking this crash bug has been fixed.

PR #549, opened 2025-01-30
Prevent null errors

### LINKED ISSUE #548: Variable References - Math - Putting in bizare crap should not crash gum
Image

Image

### LINKED ISSUE #584: InteractiveGue.CurrentInputReceiver should be nulled out when it is removed from manager.
Currently it can stick around after the screen is destroyed.

kaltinril:
Fixes #548
Resolves the crashing in #545  (Only fixes the crash, not the logic issue)
Resolves the crashing in #544  (Only fixes the crash, not the logic issue)


vchelaru:
Apologies, I didn't see this PR and I went ahead and already fixed the issue :( Sorry I'll check PRs first next time.

PR #570, opened 2025-02-02
Adjust the color display so it fills the height, but gives a 5 pixel …

kaltinril:
…border top and bottom

Relates to #468   this is item 12 on the list


kaltinril:
image

https://github.com/user-attachments/assets/4befde62-31df-452b-aad2-5034ee095649

PR #640, opened 2025-03-13
generate code button should only be visible when the other buttons ar…

### LINKED ISSUE #639: (MAJOR) Gum code - Generate Code button gone
A new project doesn't popup the 2 buttons to setup the code tab and the generate code button is missing

steps for me to reproduce:
1. open gum
2. file > new project
3. create a folder, open it, and save the new project
4. create a component
5. drop a standard element onto it
6. click on the component
7. click on the code tab
8. notice the 1 click button is not visible
9. notice that the Project-Wide Code Generation is visible
10. Notice that the Element Code Generation is visible
11. Notice that the generate code button is missing

Image

kaltinril:
…e not visible

fixes #639


vchelaru:
Thanks for looking into this. I commented on the original issue, mentioning that this behavior is as designed. Let me know if you think the behavior should be modified.

PR #644, opened 2025-03-14
643 - fix codegen for components creating un-castable GraphicalUiElem…

Seismicsentinel:
See issue - worked this out in discord


Seismicsentinel:
might be good to not eat this exception being thrown here but this fixes the immediate issue

PR #707, opened 2025-05-03
Added  an shortcut function to fetch specific screen.

niokasgami:
as explained I added an tiny shortcut function to reduce the repeating of  fetching a specific screen from a Project.

it work as the same as :
`var screen = mygumProject.Screen.find(item => item.Name = "name");`

### DIFF
--- a/GumDataTypes/GumProjectSave.cs
+++ b/GumDataTypes/GumProjectSave.cs
@@ -401,6 +401,21 @@ public class GumProjectSave
 		}
 #endif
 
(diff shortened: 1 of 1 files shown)


vchelaru:
Could you add the same for component and standard elements?

niokasgami:
sure I would need to analyse the code a little mroe to see where those are situed.

(your project is quite huge)

niokasgami:
and I added the specific two function you requested.

PR #735, opened 2025-05-08
Now when a passwordChar is set, it will instantly update the visual

### LINKED ISSUE #726: I think I should be able to change the passwordChar during runtime
I changed the password char and then called UpdateState() but it didn't update.

I then typed a single letter into the passwordbox and the displayed char's updated.

Perhaps this isn't intended behavior?  Just was trying to make it so I didn't have to add multiple password boxes for the wpf to gum testing.

Also, if I do this, the CHAR_WIDTH seems to be messed up for kerning and as such, I can click past the end of the text.

https://github.com/user-attachments/assets/4882f229-2411-4e40-afdc-ceb5d284703a

kaltinril:
Fixes #726

I validated it works in the video below. 
I made sure that the visual change didn't mess up the CaretIndex or visual position.

https://github.com/user-attachments/assets/3ecb7fa8-22ed-4b34-9b85-a65655a2c012

### DIFF
--- a/MonoGameGum/Forms/Controls/PasswordBox.cs
+++ b/MonoGameGum/Forms/Controls/PasswordBox.cs
@@ -56,7 +56,21 @@ public class PasswordBox : TextBoxBase
 
     // Update Gum's default to include this first:
     //public char PasswordChar { get; set; } = '●';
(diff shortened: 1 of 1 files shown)


vchelaru:
This is perfect, thanks!

PR #739, opened 2025-05-08
Dragging now scrolls treeview up and down if able to scroll. 

### LINKED ISSUE #270: QOL: Dragging an element to the top of the Project area should "scroll" the window up.
QOL = Quality Of Life (Improvement)

I was trying to add in a Text instance to my screen, but I couldn't get to the screen.

I know I could just collapse the Components, but if you're in the middle of a drag, I don't want to cancel and then re-start my drag/drop.

image Gum_fgwky9NFv3

kaltinril:
1. Dragging now scrolls treeview up and down if able to scroll. 
2. Node hovered over will highlight when dragging.
3. Also added in automatic node expanding on drag+hover.

Fixes #270

https://github.com/user-attachments/assets/18104321-c219-4ae4-82d1-82e7b3c9517d


vchelaru:
This is so awesome!

PR #744, opened 2025-05-09
Fixed Some Grammar, Typos, Reworded Some Stuff, and Tweaked The Manual Install Page

erievs:
My last pull request had some minor typos, so I needed to fix them.

I tweaked and reworded the auto Linux setup a bit.

I reworded the Linux prerequisites section a bit.

I added some more context to step 10 in the manual page, by explaining that BASH is what's most commonly used on Linux and macOS 10.14 and below, while ZSH is used on macOS 10.15 and higher. 

I changed it from /bin/Gum to /bin/gum, as it is more uniform across shells (on bash it'd be Gum, while fish it'd be gum).


vchelaru:
Again this is very much appreciated thanks!!

PR #746, opened 2025-05-11
GraphicsDeviceControl.DefaultRenderTarget

nkast:
this replaces Present(...) to the backbuffer with a render target


nkast:
This need work.
It shows how to draw to a rendertarget and then copy the rendertarget to a Bitmap.

PR #747, opened 2025-05-11
replace XNA with KNI and convert to AnyCPU 

nkast:


vchelaru:
I've reviewed this a bit, so far all looks like it's working great. I had to change a few minor things to address half pixel offsets and some artifacts from rendering to a render target, but now everything seems to work fine, performance is good.

PR #752, opened 2025-05-13
Project references in Samples\MVVM\MonoGameAndGum fixed

tonybierman:
Relative pathing was wrong for project references


vchelaru:
Thanks for fixing this, very much appreciated!!

PR #754, opened 2025-05-13
perf PaintRendertarget

nkast:
- dispose bitmap & renderTarget in Dispose(bool)
- set RenderTargetUsage to _PreserveContents_ as was the backbuffer before the switch to rendertargets.
there is a comment about 'IsRenderTarget containers'. @vchelaru  if something is setting it's own rt and then restore it with SetRenderTarget(null), now you'd have to pass down the  GraphicsDeviceControl.DefaultRenderTarget and change it to SetRenderTarget(defaultRenderTarget).

perf changes:
- copy pixel rows with Parallel.For(...)
- set InterpolationMode.NearestNeighbor & CompositingMode.SourceCopy for graphics.DrawImage(...)
- use Format32bppPArgb bitmap format (premultiplied) instead of Format32bppPrgb.
this effectively tells DrawImage/DrawImageUnscaled not to reapply the alpha. Good perf gains and also doesn't require us to set alpha = 255.
 - Switch  the RenderTarget SurfaceFormat to Bgra32 which is compatible with GDI/Bitmap. Copy the entire row with Marshal.Copy(...). That's possible because we no longer need to patch alpha channel.

### DIFF
--- a/XnaAndWinforms/GraphicsDeviceControl.cs
+++ b/XnaAndWinforms/GraphicsDeviceControl.cs
@@ -12,6 +12,7 @@ using System;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.Runtime.InteropServices;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using Microsoft.Xna.Framework.Graphics;
 #endregion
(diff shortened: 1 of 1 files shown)


vchelaru:
In my tests this is working well. Thanks again.

PR #791, opened 2025-05-25
fix: revert a commit that used collection expression

vicdotexe:
It looks like [this commit](https://github.com/vchelaru/Gum/commit/7fbc5f78314c15fd8f034ba908c38120b75d3d55) was removing the ^ operator (index from end) to be compatible with netstandad2.0 but also introduced the c#12 collection-expression feature. (maybe on accident as I would think this would cause similar problems?)

Reverting that part of the change since it's currently breaking KniGum and failing the pipelines.


vchelaru:
Even though I don't love using new syntax for its own sake, I think it's worth keeping this and upgrading the csproj to LangVersion 12.0.

Issue #873, opened 2025-06-10
Editor undoing changes after making one change and filewatch timer expires


violetsamber:
Context:
Linux Mint Cinnamon using a Bottles container running sys-wine-10.0 with DXVK, VK3D disabled.
I have Element Code Generation, Generation Behaviour set to NeverGenerate.

The issue I'm running into is that Gum is undoing changes in when I'm in the middle of making them. 
The undo matches up with the filewatch timer hitting zero.

https://github.com/user-attachments/assets/c9b8f246-27e1-4d66-8bea-8c252cbb643d

![Image](https://github.com/user-attachments/assets/c083f2cf-491d-4e8f-bb1f-da0c5eeaac3e)

violetsamber:
Just built with this change and tested that change on linux and its no longer causing the revert. 
The filewatcher window log does not show that it saw anything.

vchelaru:
I believe this one is now fixed.

PR #879, opened 2025-06-10
Fix linux refreshing issues

violetsamber:
Applies the fix discussed here https://github.com/vchelaru/Gum/issues/873 to the rest of the usages of `FileWatchLogic.Self.IgnoreNextChangeOn`

Changes
FileWatchLogic.Self.IgnoreNextChangeOn TO FileWatchManager.Self.IgnoreNextChangeUntil with a one second timeframe 

The one second time frame was enough in my testing.

### DIFF
--- a/Gum/Commands/FileCommands.cs
+++ b/Gum/Commands/FileCommands.cs
@@ -162,7 +162,7 @@ namespace Gum.Commands
                     }
                     else
                     {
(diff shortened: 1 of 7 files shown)


vchelaru:
Need to also try force save to see if that uses the non-time based saving.

PR #941, opened 2025-06-23
925 textruntime and text have visual issue with caret position

vchelaru:


vchelaru:
The test still fails due to being off by 1 pixel. Need to investigate why before accepting this.

PR #949, opened 2025-06-26
Add support for Numpad ENTER.  Add comments and formatting.

kaltinril:
Adds comments, but does NOT attempt to fix the keyboard repeat rate.

I could pull out lines 323 -> 336 in Keyboard.Activity
and lines 338->361 into methods to make it more readable, not sure the performance hit of adding 2 more method calls though?

323 -> 336 =>   HandleFreshKeyPress()
338 -> 361 =>   HandleRepeatedKey() or HandleKeyStillDown() or something?

I could also work on adding the WINDOWS ONLY settings detection for REPEAT rate, using SystemInformation.KeyboardSpeed; or the User32.dll if you'd like.  Just doesn't feel worth the effort?

I could also expose the 2 parameters on keyboard somehow if you'd like.


kaltinril:
I don't know how to do the repeat rate for windows only.  It requires system.windows.forms, which won't work on mac, linux, console, etc....

kaltinril:
@vchelaru  unless you have info on how to do the windows piece, you may merge this, I've pulled in main (master) and moved those 2 bits of code out into their own methods for readability.

PR #976, opened 2025-07-02
964 add visuals v2

kaltinril:
Creating this draft so as I make changes you can make comments IN LINE.

If this is not a good idea, cancel it.


MonoGameGum/Forms/FormsUtilities.cs
```diff
@@ -45,10 +46,13 @@ public static void InitializeDefaults(Game? game = null, SystemManagers? systemM
                 "You must call this method after initializing SystemManagers.Default, or you must explicitly specify a SystemsManager instance");
         }
 
-        TryAdd(typeof(Button), typeof(DefaultButtonRuntime));
+        // Get the UI Spite Sheet that we loaded in SystemManager.Initialize()
+        var uiSpriteSheetTexture2D = RenderingLibrary.Content.LoaderManager.Self.GetDisposable("UISpriteSheet.png");
```
kaltinril, line 50: I wanted to "pass this down" into the Visuals, but I don't think that's possible?

Also, the Changes in the TypeOf here are TEMPORARY, i need to change them back once I get a METHOD that allows me to change to V2.

MonoGameGum/Forms/FormsUtilities.cs
```diff
@@ -45,10 +46,13 @@ public static void InitializeDefaults(Game? game = null, SystemManagers? systemM
...
+
+        TryAdd(typeof(Button), typeof(ButtonVisual));
         TryAdd(typeof(CheckBox), typeof(DefaultCheckboxRuntime));
         TryAdd(typeof(ComboBox), typeof(DefaultComboBoxRuntime));
-        TryAdd(typeof(Label), typeof(DefaultLabelRuntime));
+        TryAdd(typeof(Label), typeof(LabelVisual));
```
kaltinril, line 55: Will be reverted once I create a switch to visuals 2 method of some kind.

MonoGameGum/Forms/FormsUtilities.cs
vchelaru: What you have here is fine for the texture. No need to pass it down.

Sounds good on the typeof changes.

RenderingLibrary/SystemManagers.cs
```diff
@@ -254,6 +254,9 @@ public void Initialize(GraphicsDevice graphicsDevice, bool fullInstantiation = f
                 LoadEmbeddedFont(graphicsDevice, "Font18Arial_Italic");
                 LoadEmbeddedFont(graphicsDevice, "Font18Arial_Italic_Bold");
 
+                var uiSpriteSheet = LoadEmberddedTexture2d(graphicsDevice, "UISpriteSheet.png");
```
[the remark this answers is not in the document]
kaltinril: I've moved the logic so that it will only load if they select V2.

Tested with WPF and WPF_v2 projects, both load still.

image

kaltinril:
https://github.com/user-attachments/assets/db91e5fd-ed13-431f-80a1-eb334dee7b91

State/status as of now

Radio buttons V2 added (mimic Gum Tool, but change to Circle in gum tool and gum code only)

kaltinril:
@vchelaru  this is ready for unit tests but I couldn't figure out how to pass in V2 and make sure it used the texture etc so I didn't try to start writing any unit tests.

If you can help me get started with a V2 test I can try to write a bunch.

vchelaru:
@kaltinril I just pushed a change that includes a unit test file for each of the visual types. Notice that I already created a unit test where the type is instantiated. Be aware that unit tests do not have a graphics device so they cannot create a Texture2D. Therefore, any tests where you are depending specifically on a Texture2D existing will not work. 

If we want to write tests that require a graphics device, this would probably be a bigger project because I don't know how to do it off the top of my head.

kaltinril:
@vchelaru  I don't think there is much left, I've changed this to a real PR.  If you think there is other stuff still outstanding I'm forgetting about, please let me know and change it back to a draft.

Issue #1053, opened 2025-07-14
Add SonarScanner github action?


kaltinril:
If there is any value you see in the SonarQube SonarScanner results, we can have it auto scan via github actions with instructions link below.

However, if you feel the information it provides is useless for this project, cancel/close this ticket.

https://docs.sonarsource.com/sonarqube-server/10.7/devops-platform-integration/github-integration/adding-analysis-to-github-actions-workflow/

You can see the scan results from the last time I ran it here:

https://sonarcloud.io/summary/new_code?id=kaltinril_gum&branch=master

kaltinril:
<img width="1220" height="788" alt="Image" src="https://github.com/user-attachments/assets/94f722bb-4960-4356-a1ae-5c2757375075" />

<img width="395" height="1206" alt="Image" src="https://github.com/user-attachments/assets/892e27e0-09ca-4f12-8e0d-951b666821c8" />

kaltinril:
Branch: add-sonarscanner

https://github.com/vchelaru/Gum/tree/add-sonarscanner

kaltinril:
Also, you should probably setup your own sonarcloud free account and we should change the KEY and TOKEN and URL to yours :) 

Just so you own it since you own gum

PR #1080, opened 2025-07-18
Create sonarscanner.yml

kaltinril:
I'm leaving 2 secrets I need you to create @vchelaru  in discord DM.

secrets.SONAR_HOST_URL
and
secrets.SONAR_TOKEN

### DIFF
--- a/.github/workflows/sonarscanner.yml
+++ b/.github/workflows/sonarscanner.yml
@@ -1,7 +1,7 @@
 # This workflow will build a .NET project
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-net
 
-name: .NET
+name: .SonarQube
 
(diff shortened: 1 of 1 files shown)


kaltinril:
This relates to #1053

kaltinril:
ok @vchelaru it works, it takes about 4 minutes for it to run the build and analysis scan and upload.
The URL ends up being obfuscated so I think I'm going to hardcode that instead.  This obfuscation makes it hard to go to the scanned result.  It's also hard to find in the logs though, so perhaps it doesn't matter and we just manually know to go to the sonarcloud.io site?

I also want to add in the .net core testing project MonogameGum, I'll add that now.

image
image

kaltinril:
From the logs from the run, the 154 tests did execute.

image

PR #1215, opened 2025-08-07
Add missing namespace that was causing generated code compiler errors regardless of `using` directives due to `global::` prefix

Pituivan:
Namespace was missing in two specific lines which caused compiler errors because of the use of `global::` prefix, or if their proper namespace were missing.

image


CodeOutputPlugin/Manager/CodeGenerator.cs
```diff
@@ -1006,14 +1006,14 @@ private static void AddFindByNameAssignment(CodeGenerationContext context)
             {
                 context.StringBuilder.AppendLine(
                     $"{context.Tabs}{context.InstanceNameInCode} = this.Visual?.GetGraphicalUiElementByName(\"{context.Instance.Name}\") as " +
-                    $"global::{GetClassNameForType(context.Instance.BaseType, context.VisualApi, context)};");
+                    $"global::MonoGameGum.GueDeriving{GetClassNameForType(context.Instance.BaseType, context.VisualApi, context)};");
```
vchelaru, line 1009: What happens if the instance's base type is a component? Wouldn't that cause this to generate invalid code?

vchelaru:
I added a new project in the latest commit. If you codegen from it, it will be a good test of the codegen system. As of right now, on this branch, it does not compile:

image

PR #1220, opened 2025-08-09
1204/tag deleted component instances as missing (CodeGen)

### LINKED ISSUE #1204: Code Generation should somehow tag instances that are referencing a missing Component as missing
When deleting a Component, instances of it will be generated. When doing Forms-first code generation, since the component is not found it is not considered a Forms component and it does not generate with the appended "Runtime".

This can cause confusion as the user may wonder why the type was changed, rather than focusing on the real problem - that the component was deleted.

Therefore, the code generation should append a comment or somehow indicate "Could not find type XXXXX".

Pituivan:
Replaced definitions of instances that reference missing components with a warning comment explaining that their type could not be found.

Also removed the part of the code where these instances were initialized.

Resolves #1204

- - -

I think this is a better approach than leaving a comment next to the buggy instances of missing components, since now if the instance is invalid you will just not be able to use it in code. However, I may be missing some things I haven't taken into account. What do you think @vchelaru?


Pituivan:
Now deleted component instances are excluded from instance initialization.

| image |
|:--:|
| `DeletedComponentInstanceName` initialization isn't in this method anymore |

Pituivan:
Tested it and test projects compile with no problem

PR #1224, opened 2025-08-09
1221: Ignore spaces in element and variable name verification

Pituivan:


Pituivan:
Closed this branch to separate it into two different fixes (preventing duplicate names by standarizing and treat spaces as underscore in code generation)

PR #1236, opened 2025-08-10
Create MVVM infrastructure for tab management

vicdotexe:
- Create PluginTabContainerViewModel for managing collections of tabs
- Refactor PluginTab to be MVVM-based with data binding support:
  - Add properties for IsVisible, IsSelected, ParentContainer
  - Remove dependency on GuiCommands and WPF TabItem
  - Add RelayCommand for Hide operation
- Create MainPanelViewModel implementing ITabManager:
  - Manages all tab containers for different locations
  - Handles tab creation, removal, and visibility
- Update service registration to use MainPanelViewModel as ITabManager
- Add factory method support for dependency injection
- Update MainPanelControl constructor to accept ViewModel


Gum/Plugins/InternalPlugins/Behaviors/MainBehaviorsPlugin.cs
```diff
@@ -219,7 +219,7 @@ private void UpdateTabPresence()
         {
             viewModel.UpdateTo(asComponent);
 
-            this.behaviorsTab.Show(focus: false);
+            this.behaviorsTab.Show(false);
```
vchelaru, line 222: "magic bool" params should either have local variables or parameter (like `focus:`) to explain their usage.

Issue #1221, opened 2025-08-09
CodeGen doesn't care about spaces in C# class names


Pituivan:
Gum doesn't have problem with letting you use spaces in element names. But C# does. And CodeGen doesn't do anything about it.
<img width="600" height="118" alt="Image" src="https://github.com/user-attachments/assets/1e4105ee-7c51-43e9-b668-7fb02fe76b6e" />

This is not as trivial as just removing the spaces or replacing them with underscores, since if we did that, we'd be risking duplicate names: `MyElement` and `My Element`, or `My_Element` and `My Element`, would both translate to a C# class with the same name, causing compiler errors.

To solve this: 
- [x] Make `NameVerifier` and the code that verifies variables treat spaces as underscores
- [x] Make `CodeGenerator` replace spaces with underscores

kaltinril:
Assigned to @Pituivan to learn and work.

PR #1246, opened 2025-08-11
1221 treat spaces as underscores in category name verification

Pituivan:
Closes the first part of #1221 as finished (this time for real) — in PR #1232 I had forgot to handle also category names in name verification.

Before my changes, you could create a state category called `MyCategory` and another state category called `My Category` in the same element.

However, taking into account that we will replace spaces with underscores in Code Generation, that would cause name conflicts.

So now category name verification also treats spaces as underscores so we can safely implement that in CodeGen,

### DIFF
--- a/Gum/Dialogs/AddCategoryDialogViewModel.cs
+++ b/Gum/Dialogs/AddCategoryDialogViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using Gum.DataTypes;
 using Gum.DataTypes.Variables;
+using Gum.Managers;
 using Gum.Services.Dialogs;
 using Gum.ToolCommands;
 using Gum.ToolStates;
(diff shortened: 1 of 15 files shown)


Gum/Dialogs/AddCategoryDialogViewModel.cs
```diff
@@ -40,13 +44,8 @@ protected override void OnAffirmative()
...
     protected override string? Validate(string? value)
     {
-        
-        if (StateContainer is ElementSave element &&
-            element.GetStateSaveCategoryRecursively(value, out ElementSave categoryContainer) is not null)
-        {
-            return
-                $"Cannot add category - a category with the name {value} is already defined in {categoryContainer}";
-        }
+        if (StateContainer is ElementSave element && _nameVerifier.IsCategoryNameValid(value, element, out string whyNotValid))
```
vchelaru, line 47: Shouldn't this be `_nameVerifier.IsCategoryNameValid(...) == false` ?

Gum/Managers/NameVerifier.cs
```diff
@@ -79,6 +79,25 @@ public bool IsElementNameValid(string componentNameWithoutFolder, string folderN
...
+                if (Standardize(item.Name) == standardizedName)
+                {
+                    existingName = item.Name;
+                    return true;
+                }
+
+                return false;
+            });
+
+            whyNotValid = $"Cannot add category — a category with the name {existingName} is already defined in {categoryContainer}";
```
vchelaru, line 98: This should probably only assign whyNotValid if there is an existing item. Otherwise this may cause confusion when debugging.

PR #1248, opened 2025-08-11
Replace spaces with underscores in CodeGen

### LINKED ISSUE #1221: CodeGen doesn't care about spaces in C# class names

Pituivan:
Closes #1221

### DIFF
--- a/CodeOutputPlugin/Manager/CodeGenerator.cs
+++ b/CodeOutputPlugin/Manager/CodeGenerator.cs
@@ -385,6 +385,13 @@ public class CodeGenerator
 
         context.StringBuilder.AppendLine(context.Tabs + header);
     }
+
+    private static string ToCSharpName(string nonFormatted)
+    {
+        return nonFormatted
+            .Replace(".", string.Empty)
+            .Replace(" ", "_");
+    }
     
     /// <returns>
     /// The corresponding class name, or <c>null</c> if that type couldn't be found.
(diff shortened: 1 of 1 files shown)


Pituivan:
I think it would be easier to simply store the variable names with underscores instead of spaces. That'd be tidier and I wouldn't have to touch `CodeGenerator`.

Pituivan:
Gonna sleep now.

My initial idea on how to do this was the following:
- First, either make `NameVerifier.Standardize` a public static method, or rather move it to somewhere where it's globally accessible or at least accessible from InternalPlugins.
- Secondly, add a `StandardizedName` property to VariableSaves, ElementSaves, etc.
- Finally, make `CodeGenerator` use StandardizedName instead of just Name property.

However I'm not as convinced it's a great idea. Maybe it should be better to make CodeGen standardize names on the fly.

I've gotta ask @vchelaru who knows the architecture the best. Do you think it would be any beneficial to have a `StandardizedName` property in the mentioned classes, or should I rather use normal name and standardize it from CodeGen directly?

Pituivan:
After trying it out in the test it seemed to work... But test projects already give me compiler errors by default. At a quick glance, everything seems to be alright, although can't be 100% sure until getting the test projects to work without errors for me.

But overall it looked like my changes worked just fine.

PR #1261, opened 2025-08-12
Refactors plugin tab management

### LINKED ISSUE #1249: Code tab should not be first tab in bottom section

vicdotexe:
This change refactors how plugins are managed and displayed in the Gum editor. The previous implementation used multiple `PluginTabContainerViewModel` objects which overcomplicated the process. This commit introduces a single collection of `PluginTab`objects, with collection views for each control, simplifying the overall architecture.

- Fixes some issues with tab visibility and selection.
- The ITabManager interface has been simplified by removing the RemoveControl and ShowTabForControl methods since the tabs can be shown/hidden via the PluginTab itself.
- fixes #1249 -- this doesn't change the code-tab from being first, however it addresses the underlying issue that the code-tab becomes selected by default


PerformanceMeasurementPlugin/MainPlugin.cs
```diff
@@ -46,7 +46,7 @@ public override void StartUp()
 
         public override bool ShutDown(Gum.Plugins.PluginShutDownReason shutDownReason)
         {
-            _tabManager.RemoveControl(view);
+            //_tabManager.HideTabForControl(view);
```
vchelaru, line 49: How are tabs removed when a plugin dies?

vicdotexe: This is that plugin we spoke of in DM that isn't even added or used.

PR #1263, opened 2025-08-13
1262 possible name conflicts in states and categories

### LINKED ISSUE #1262: Possible name conflicts in states and categories
You should not be able to create a Category in a Element which name is the same as the Element's. And neither should you be able to create a State named the same as the Category it was created in.

This would cause name conflicts (compiler errors) in CodeGen. I think the cleanest solution is not to allow this at all.

Image

Pituivan:
Resolves #1262


Pituivan:
Damn it, tests are failing 😥

Issue #1270, opened 2025-08-13
Texture Left in Texture Address not looping when set to negative value


Yuan-Zzzz:
When setting Texture Left to a positive value, the texture scrolls/loops as expected.
However, when setting Texture Left to a negative value, the texture does not loop on the left side.

<img width="1733" height="958" alt="Image" src="https://github.com/user-attachments/assets/3ee3ae05-7b86-4356-8a86-c71592a98f23" />

<img width="1966" height="700" alt="Image" src="https://github.com/user-attachments/assets/c683d0b3-1e52-41ae-8cd2-1967aa2e22ef" />

vchelaru:
Fixed

PR #1275, opened 2025-08-14
Fix texture looping with negative Texture Left

Yuan-Zzzz:
Integer division in C# truncates toward zero, so negative texture offsets produce wrong tile indices, breaking wrap on the left/top.


Yuan-Zzzz:
#1270

vchelaru:
Works great in my test:

https://github.com/user-attachments/assets/fd11393f-523c-4331-b1da-4b6c0af07722

PR #1288, opened 2025-08-15
Fixed issue with missing usings and a method signature

profexorgeek:
This PR addresses compile issues on my machine. Some linq and Func calls needed using statements and the `Split` method would not compile without passing an explicit `options` parameter. Intellisense showed that an overload exists with the correct signature and yet it wouldn't compile.

I don't understand why this is the case but this PR fixes it in a way that should be "safe" by just making the second arg explicit.


profexorgeek:
We had weird discrepancies between builds on different machines. After a deep clean, the problem that triggered this fix went away 😳 and so this PR should be closed without a merge.

PR #1305, opened 2025-08-19
Improved single-page rendering of PNGs for outlined fonts

vchelaru:
[BREAKING] TextRuntime now uses XNA blend states instead of abstract Gum blend states. Default Gum blend states now have the blend factor and write channels enabled.


vchelaru:
I don’t think so, it’s isolated to monogame

PR #1355, opened 2025-08-30
195 Include children when creating component

### LINKED ISSUE #195: Create components should recursively include all instances
1. Create a container in a screen
2. Add a Text as a child to the container
3. Add a Sprite as a child to the container
4. Select the container
5. Right-click and select to create a component
6. Enter a name and click OK

Observe - only the container is added
Expected - all children should be added recursively

Pituivan:
Resolves #195 

https://github.com/user-attachments/assets/33f4330d-f8a9-4bdf-b686-e875254cb276


Pituivan:
Accidentally closed it

Gum/PropertyGridHelpers/VariableInCategoryPropagationLogic.cs
```diff
@@ -3,28 +3,32 @@
...
-        IFileCommands fileCommands)
+
+    public VariableInCategoryPropagationLogic()
     {
-        _undoManager = undoManager;
-        _guiCommands = guiCommands;
-        _fileCommands = fileCommands;
+        _undoManager = Locator.GetRequiredService<IUndoManager>();
+        _guiCommands = Locator.GetRequiredService<IGuiCommands>();
+        _fileCommands = Locator.GetRequiredService<IFileCommands>();
```
vicdotexe, line 29: Why are we moving these already injected dependencies to be acquired from the Locator?

Tool/Tests/GumToolUnitTests/Commands/EditCommandsTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        var mock = new Mock<IDialogService>();
+        mock
+            .Setup(x => x.GetUserString(It.IsAny<string>(), "Create Component from selection", It.IsAny<GetUserStringOptions>()))
+            .Returns("ComponentName");
+
+        return mock;
+    }
+    
+    [Fact]
+    public void ShowCreateComponentFromInstancesDialog_ShouldIncludeChildren()
```
vicdotexe, line 85: Isolating the setup and assertion logic outside of this test method makes it hard to understand what's happening here. If there were some repeated logic going on here it might be helpful in keeping things dry, but as it is now, I think it's doing more harm than good.

Gum/ToolCommands/ProjectCommands.cs
```diff
@@ -15,11 +15,11 @@
 
 namespace Gum.ToolCommands;
 
-public class ProjectCommands
+internal class ProjectCommands : IProjectCommands
 {
     #region Fields
 
-    static ProjectCommands mSelf;
+    static IProjectCommands mSelf;
```
vicdotexe, line 22: Can you double check if this is actually used? We shouldn't have any static singletons being added to the the container (with the exception of ObjectFinder).

Gum/PropertyGridHelpers/VariableInCategoryPropagationLogic.cs
Pituivan: My bad, didn't notice that. Just found out in this PR that Rider has an automatic refactor feature of replacing use of a class with its interface when possible, so didn't actually read that part. I'm fixing it

Pituivan:
Solved

vchelaru:
Closed per @Pituivan 's request

PR #1359, opened 2025-09-01
130 Extract component from instance + `Create Component` rework

### LINKED ISSUE #130: [Feature Request] Allow making a system wide component from an existing screen element

Pituivan:
### [In progress]

This PR branch is a sub-branch of the branch from #1355, so you'll see its history along with the specific history of this PR. Since we're squash-merging, that won't change when #1335 is merged to master.

Resolves #130.

## Changes:
- Replaced `Create Component` context menu actions two other more specific ones:
  - `Create Component from selection`: Only appears when more than one instance is selected. Behaves just like old `Create Component`.
  - `Create Component from this`: Only appears when a single instance is selected. Behaves similar to old `Create Component` but **does not create the component as a container of the selected instance**; instead it creates a component from the selected instance.
    > I'm not 100% sure of this one. What do you think about it? I think it's better this way but I need your opinion.
- Added a new context menu action: `Extract Component`. Based on the request from [#130](https://github.com/vchelaru/Gum/issues/130), it behaves exactly the same as `Create Component from this` action but also replaces the selected instance with an instance of the newly created component preserving the name.


vchelaru:
Closed per @Pituivan 's request

PR #1583, opened 2025-10-17
1547 enhance slider click

### LINKED ISSUE #1547: Clicking in slider should JUMP to click location
When I click i the slider in a color it moves by 1 value (255 -> 254)

Should move to exact location, or, jump a larger value when clicking on the BAR.

https://github.com/user-attachments/assets/e51de209-274e-4a19-9f04-4e65e3fa4bcd

kaltinril:
Made clicks jump to the value.
Fixed missing tick marks in GUM (and FRB, see PR over there)

Fixes #1547 

https://github.com/user-attachments/assets/1c6c9eef-7e21-44ae-b831-fc290f7d80b6

### DIFF
--- a/WpfDataUi/Controls/SliderDisplay.xaml.cs
+++ b/WpfDataUi/Controls/SliderDisplay.xaml.cs
@@ -188,22 +188,35 @@ namespace WpfDataUi.Controls
             MinValueText.Text = (this.MinValue * DisplayedValueMultiplier).ToString();
             MaxValueText.Text = (this.MaxValue * DisplayedValueMultiplier).ToString();
 
-            // Ticks are only visible if the styles defined in these below files
-            // Contain a Slider with at least 1 TickBar entry
-            // ..\Gum\Gum\Themes\Frb.Styles.Defaults.xaml
-            // ..\FlatRedBall\FRBDK\Glue\Glue\Themes\Frb.Styles.xaml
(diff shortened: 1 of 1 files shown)


kaltinril:
> Can we fix the mis-alignment between 0 and the first tick? This feels wrong to me. image

maybe?

kaltinril:
I was able to fix the color after converting to WPF

image

vicdotexe:
@vchelaru @kaltinril  It does look a bit odd at first glance, just because it's a single digit, but it wouldn't really make sense to center the text value on that tick the way you think it would... if you centered it on the tick, and the values had say 3 digits, it would end up making the slider bar smaller. I think it's accentuated because 255 at that font size just happens to perfectly center on the end tick, so it makes the 0 look off.

Imagine sliders in the same column like that being stretched horizontally, each with different digit count min's: 0, 10, 100, 1000. You'd end up with with a stack of sliders who's bars didn't line up.
image

@kaltinril I wouldn't muck with the base implicit style to conform it to this spot. displaying a minmax should probably still live outside of the base style/template, because we don't want to make assumptions: string formatting, is it time, whole numbers, percentage, decimals, how many digits, should it be on top or below or outside edges, colors, icons (mute vs speaker), do we even want to display the values, etc. you'll want to do any scenario specific closer to that.

kaltinril:
Ok waiting for direction from @vchelaru  on this then

PR #1590, opened 2025-10-18
Check for missing filename and exit without error, if saving failed s…

### LINKED ISSUE #1548: Escaping the new project creation crashes gum (and should not)
Steps

1. Click New Project
2. Press ESC or click cancel
3. gum crashes

I suspect this is because there is no check for null on the selected file path.

https://github.com/user-attachments/assets/9648764c-3440-48e6-9d74-3fdcc8aeb696

kaltinril:
…pit out a failure

Fixes #1548


kaltinril:
@vchelaru was the animation fix for something that I broke?i get confused when you add things to my pr

PR #1610, opened 2025-10-19
1558 resizing window should be done from the outside rather than inside

### LINKED ISSUE #1558: Resizing window should be done from the "outside" rather than "inside"
This should behave the way normal Windows works, where resizing happens outside of the control.

This should happen by modifying V2 controls to:

1. Have its resize containers be positioned outside
2. Have the window check for clicks outside of its bounds

kaltinril:
Fixes #1558


kaltinril:
Be aware though, if the window is inside a panel, then the drag action can't trigger if the border object is outside the bounds of the invisible panel (like my in WPF project)

https://github.com/user-attachments/assets/3f9f7bb7-f0ca-4c21-bffe-eb6fd749af5d

Issue #1612, opened 2025-10-19
Research how to save Gum tool artifact


vchelaru:
This task is partly a research task - we are already building Gum on Windows as part of PR checks to make sure it compiles, but the artifact isn't saved anywhere.

We need to figure out how to build Gum. Specifically, what we want is:

1. Set the version number to Year.Month.Date
2. Build the entire Gum.sln in either debug or release - verify that the plugin folder gets populated with all of the plugin .dlls - you can compare with what you have on your machine to see what should be there
3. Save the resulting bin/debug or bin/release built file as a .zip (or anything else if it's easier to do so)
4. Figure out how to use these artifacts in a release - do they have permalinks? 

After learning how to do this, let's discuss whether we want to update our release process to use this instead of manual builds. Perhaps this will address the false positives (or true positives?) of malware in our zip files?

kaltinril:
@vchelaru 
Ok I got a branch with a workflow script.

Looks like this has to be merged into main before I can test the action though...  

It should PROMPT you to select TEST which will sent the build as prerelease and draft.

If it works, we should technically be able to merge this and your nuget deploy into 1 single thing if you'd like.

I'll create a PR to master so this file can get there so I can test it if you'd like.  If not, let me know how you want to proceed

PR #1624, opened 2025-10-24
Test workflow that will build, package, tag, and release

kaltinril:
See message here:
https://github.com/vchelaru/Gum/issues/1612#issuecomment-3441501696

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -10,10 +10,6 @@ on:
         options: [release, prerelease, test]   # test = draft
         default: test
         required: true
-      notes:
-        description: 'Optional extra notes'
-        type: string
-        required: false
 
 permissions:
   contents: write
@@ -31,6 +27,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
+          ref: master
           submodules: recursive
           #fetch-depth: 0
 


.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,98 @@
...
+          $tag   = "${prefix}_" + (Get-Date).ToString('MMMM_dd_yyyy')
+          if ($kind -eq 'test') { $tag = "$tag-$($env:GITHUB_RUN_NUMBER)" }
+          "tag=$tag"               >> $env:GITHUB_OUTPUT
+          "title=$title"           >> $env:GITHUB_OUTPUT
+          "prerelease=$prerelease" >> $env:GITHUB_OUTPUT
+          "draft=$draft"           >> $env:GITHUB_OUTPUT
+
+      # 3) Create and push the annotated tag
+      - name: Create & push tag
+        if: github.event.inputs.release_kind == 'release'
```
kaltinril: We can always change this, but this supposedly will auto tag with the tag created above.  It will also be useful in the softprops which creates the release page for you with the packaged zip file

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,98 @@
...
+          git tag -a "${{ steps.meta.outputs.tag }}" -m "${{ steps.meta.outputs.title }}"
+          git push origin "${{ steps.meta.outputs.tag }}"
+
+      # 4) Create GitHub Release and attach the zip
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ steps.meta.outputs.tag }}
+          name:     ${{ steps.meta.outputs.title }}
+          generate_release_notes: true
```
kaltinril: This will automatically create the release history based on the changes since the last TAG.  This means that if we want it to work correctly we will need to only use TAGs for the RELEASES I believe unless there are more options for the softprops/action-gh-release as I'm not an expert

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,98 @@
...
+      # 2) Compute tag + title depending on the kind of release
+      - name: Compute tag + title
+        id: meta
+        shell: pwsh
+        run: |
+          $kind = "${{ github.event.inputs.release_kind }}"
+          switch ($kind) {
+            'release'    { $prefix='Release';    $prerelease='false'; $draft='false' }
+            'prerelease' { $prefix='PreRelease'; $prerelease='true';  $draft='false' }
+            'test'       { $prefix='Test';       $prerelease='true';  $draft='true'  }
```
kaltinril: I'm hoping that this will make it so when selecting test, it goes to a hidden page only you (and i?) can access.

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,95 @@
...
+    inputs:
+      release_kind:
+        description: 'Type'
+        type: choice
+        options: [release, prerelease, test]   # test = draft
+        default: test
+        required: true
+
+permissions:
+  contents: write
```
kaltinril: this is supposedly needed to TAG the repo.

PR #1647, opened 2025-10-27
Put the functionality into the XAML instead of CS

kaltinril:
Here is the fix, I don't really know what I'm doing so if there are bad things in this please fix them

### DIFF
--- a/Gum/MainWindow.xaml.cs
+++ b/Gum/MainWindow.xaml.cs
@@ -6,11 +6,7 @@ using Gum.Services;
 using Gum.Settings;
 using Gum.ViewModels;
 using System;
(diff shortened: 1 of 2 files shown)


Gum/MainWindow.xaml.cs
```diff
@@ -83,65 +79,3 @@ private void OnCloseButtonClick(object sender, RoutedEventArgs e)
 public record CloseMainWindowMessage;
 public record ThemeChangedMessage(IEffectiveThemeSettings settings);
 
-
-// Need this to respect taskbar in maximized state while using custom window chrome
```
vicdotexe: Read the comment

PR #1648, opened 2025-10-27
address fussy windowchrome

vicdotexe:
- replaces previous in-house window chrome fixes with a dedicated 3rd party control to handle all the fussy behavior when rolling custom window chrome


kaltinril:
Added approval since I tested and it resolved it for me

PR #1702, opened 2025-11-06
bug: maximized gap in multi-monitor setup

### LINKED ISSUE #1682: Maximized GUM has 3-5 pixel border that shows other apps behind it
Image

vicdotexe:
fixes #1682


vchelaru:
I verified that it was fixed for me when running in multi-monitor mode

PR #1707, opened 2025-11-08
Fix styling errors on missing project

vchelaru:
* Fixes styling error when showing popup if the previous project is missing
* Fixes popup not showing correctly because the main window hasn't loaded when trying to show missing project popup
* Fixes bug where Gum tries to load the last project every time Gum launches instead of only one time

### DIFF
--- a/Gum/Themes/FrbTabControl.cs
+++ b/Gum/Themes/FrbTabControl.cs
@@ -40,15 +40,7 @@ public class FrbTabControl
 
     public static double? GetMaxTabWidth(TabControl obj)
     {
-        var valueAsObjct = obj.GetValue(MaxTabWidthProperty); 
-        if(valueAsObjct is double asDouble)
-        {
-            return asDouble;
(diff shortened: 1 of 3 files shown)


Gum/Themes/FrbTabControl.cs
```diff
@@ -40,7 +40,15 @@ public static void SetTruncationIcon(TabItem obj, PackIconKind? value)
...
-        return (double?)obj.GetValue(MaxTabWidthProperty);
+        var valueAsObjct = obj.GetValue(MaxTabWidthProperty); 
+        if(valueAsObjct is double asDouble)
+        {
+            return asDouble;
+        }
+        else
+        {
+            return null;
+        }
```
vicdotexe: This might be left over? Might suggest reverting back to original as MaxTabWidthProperty is type Nullable<double>, it should only ever return double/null and the redundancy implies otherwise.

PR #1992, opened 2026-01-13
Add htale sample

kaltinril:
Draft only right now


kaltinril:
I've converted this to a real PR @vchelaru this contains my development version.

I've got a similar but different version in my personal github that if you want we can eventually put here instead

PR #2131, opened 2026-02-10
Add hotkey navigation from any area of the app to change the selected…

### LINKED ISSUE #2102: Next/Previous State Hotkey
When the application focus is on the variables tab, changing the state requires using the mouse to change focus to the states panel. I would like a dedicated set of hotkeys for navigating my current state within the current category, regardless of the current focus. Page Up and Page Down seem like reasonable candidates. Maybe CTRL is a modifier key that makes them navigate categories instead.

kaltinril:
… state in the category

Fixes #2102


kaltinril:
https://github.com/user-attachments/assets/1dc2fa16-c0bc-47a8-a9cf-58c3964dbbbb

PR #2199, opened 2026-02-17
2038/add vulcan checks

### LINKED ISSUE #2038: Installing GUM Tool on Fedora 43
I've tried installing Gum Tool on Fedora 43 with the install script, the manual process, and via Bottles. 

All of them result in some flavor of the following error when I launch gum:
```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at System.Runtime.InteropServices.Marshal.Release(IntPtr)
   at SharpDX.ComObject.SharpDX.IUnknown.Release()
   at SharpDX.ComObject.Dispose(Boolean)
   at SharpDX.DisposeBase.CheckAndDispose(Boolean)
   at SharpDX.DisposeBase.Dispose()
```

Is this something anyone has encountered before?

Here is the full output of when I launch GUM with wine: [gum_output.txt](https://github.com/user-attachments/files/24848618/gum_output.txt)

kaltinril:
fixes #2038 

this will install vulcan support if it's detected as available
it will also default to that

the use can revert by running: `gum d3d`
or migrate back by running: `gum dvxk`
they can still upgrade by running: `gum upgrade`

I also passed the quiet flag, so the dotnet 8 now installed without popups

Docs need to be updated still though

### DIFF
--- a/setup_gum_linux.sh
+++ b/setup_gum_linux.sh
@@ -382,28 +382,23 @@ chmod +x ~/bin/gum &> /dev/null
 
 ################################################################################
 ### Check if the bin directory is in PATH based on the shell being used
-### If not, add it to PATH and reload the shell configuration.
+### If not, add it to PATH for future terminal sessions.
 ################################################################################
 if [[ $SHELL == *"bash"* ]]; then
     if ! grep -q 'export PATH="$HOME/bin:$PATH"' ~/.bashrc 2>/dev/null; then
-        echo "Adding ~/bin to PATH in ~/.bashrc, please wait..."
+        echo "Adding ~/bin to PATH in ~/.bashrc..."
         echo 'export PATH="$HOME/bin:$PATH"' >> ~/.bashrc
     fi
-    echo "Reloading ~/.bashrc, please wait..."
-    source ~/.bashrc &> /dev/null
 elif [[ $SHELL == *"zsh"* ]]; then
     if ! grep -q 'export PATH="$HOME/bin:$PATH"' ~/.zshrc 2>/dev/null; then
-        echo "Adding ~/bin to PATH in ~/.zshrc, please wait..."
+        echo "Adding ~/bin to PATH in ~/.zshrc..."
         echo 'export PATH="$HOME/bin:$PATH"' >> ~/.zshrc
     fi
-    echo "Reloading ~/.zshrc, please wait..."
-    source ~/.zshrc &> /dev/null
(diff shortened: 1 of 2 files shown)


kaltinril:
screenshots from running the commands

Screenshot from 2026-02-16 16-01-43

Screenshot from 2026-02-16 15-58-30

screenshot in gum on linux

Screenshot from 2026-02-16 16-02-00

screenshot installing gum  (with prompt, prompt is no longer in script)
Screenshot from 2026-02-16 15-36-32

kaltinril:
converted to draft

kaltinril:
Ok, I tested this pretty good on my ubuntu 24.04.  Apparently I had messed up drivers and after removing an old kernal folder, i finally got the nvidia drivers correctly installed.



then i tested d3d first and it worked

then I switched to dxvk and it didn't, but that was because my os was messed up still and I had to install 

so i had to
purge 6.17 kernal that was old
reinstall nvidia-driver-535
install libvulkan1:i386

after that, it worked.



I'm going to make it install dxvk but default to d3d and users can manually switch to dxvk via



gum dxvk

then run gum with

gum

kaltinril:
Going to make a few more small changes first

setup_gum_linux.sh
```diff
@@ -284,11 +257,37 @@ rm -f "$GUM_ZIP_FILE" \
...
+echo "Reverting to WineD3D as the default renderer..."
+run_and_write_log WINEPREFIX="$GUM_WINE_PREFIX_PATH" winetricks d3d8=builtin d3d9=builtin d3d10=builtin d3d10_1=builtin d3d10core=builtin d3d11=builtin dxgi=builtin
+echo " - DXVK installed. WineD3D is the default. To switch: ~/bin/gum dxvk"
+
 ################################################################################
 ### Define the script variables
 ################################################################################
 echo "Creating gum script and adding to path"
-GUM_EXE_PATH=$(find "$GUM_WINE_EXTRACT_DIR" -name "Gum.exe" -type f)
+GUM_EXE_PATH=$(find "$GUM_WINE_EXTRACT_DIR" -name "Gum.exe" -type f | head -n1)
```
kaltinril: In the, VERY unlikely event there are 2 in the folder

kaltinril:
OK, this is ready, MAC still doesn't work on my non apple silicon device and I don't have an apple silicon to test. 

However I split the mac code out so that we don't have to keep trying to pay attention to it, since it's already not working.

PR #2213, opened 2026-02-18
Fix readonly source bindings without exception-driven setter creation

### LINKED ISSUE #2095: Binding should not raise exceptions if they are intended to be caught when binding to readonly ViewModel property
If a UI property (such as Label.Text) with both a get and set is bound to a ViewModel property that only has a get, there is an internal exception that is thrown. However, that exception is immediately caught and one-way binding is effectively created.

Most of the time this is okay but it does have a few problems:

1. Exceptions can be a little expensive, especially if this is a hot path, such as if bound on a large number of items (ListBoxItems).
2. Users may catch this type of exception in their IDE settings and be confused about why this is happening.

(2) is what happened here in discord:

https://discord.com/channels/586997072373481494/1073643744580079687/1468636290596606097

I had to waste time reproducing it only to find that it is caught internally. Before I realized this I assumed this was causing a crash for the user.

This is some example code:

```cs
class MyViewModel
{
    public string LabelText => "Hello";
}

protected override void Initialize()
{
    GumUI.Initialize(this, Gum.Forms.DefaultVisualsVersion.Newest);

    Label label = new();
    label.AddToRoot();
    label.BindingContext = new MyViewModel();
    label.SetBinding(nameof(label.Text), nameof(MyViewModel.LabelText));

    base.Initialize();
}
```

The exception is ultimately thrown here:

```
This exception was originally thrown at this call stack:
    System.Linq.Expressions.Expression.RequiresCanWrite(System.Linq.Expressions.Expression, string)
    System.Linq.Expressions.Expression.Assign(System.Linq.Expressions.Expression, System.Linq.Expressions.Expression)
    Gum.Forms.Data.BinderHelpers.BuildSetter(System.Type, string) in BinderHelpers.cs
    Gum.Forms.Data.UntypedBindingExpression.AttachToSource(object) in UntypedBindingExpressionBase.cs
```

This exception is caught and the binding is made one-way. I think a better approach is to check if things are writable and only attempt to create a setter if the VM property is writable.

I do realize that the exception and catch are controlling the logic flow and ultimately it works but it causes confusion for the users which requires support time.

Chevalier12:
Fixes #2095

## Problem
When binding a UI property to a readonly ViewModel property, the binding setup always attempted to build a source setter and relied on an exception to fall back to one-way behavior.

This caused:
- avoidable exception overhead in hot paths,
- debugger noise/confusion for users who break on thrown exceptions.

## What Changed
- Added `BinderHelpers.CanWritePath(Type, string)` to pre-check whether a source path is writable.
- Updated `UntypedBindingExpression.AttachToSource` to only attempt setter generation when:
  - mode is `TwoWay` or `OneWayToSource`, and
  - `CanWritePath(...)` returns true.
- Kept a narrow guarded catch around `BuildSetter` only for edge-case expression failures, not normal readonly behavior.

## Tests Added
- `BinderHelpersTests`
  - writable and readonly direct properties
  - writable and readonly nested properties
- `FrameworkElementBindingTests`
  - `ModeTwoWay_ReadonlySourceProperty_DoesNotUpdateSource`
  - `ModeOneWayToSource_ReadonlySourceProperty_DoesNotThrowOrUpdateSource`

## Validation
Ran with `DOTNET_ROLL_FORWARD=Major` and net8 override in this environment:
- targeted regression tests: passed (7/7)
- binding-focused suite: passed (52/52)
- full `MonoGameGum.Tests`: passed (502), skipped (1), failed (0)

### DIFF
--- a/MonoGameGum.Tests/Forms/Data/BinderHelpersTests.cs
+++ b/MonoGameGum.Tests/Forms/Data/BinderHelpersTests.cs
@@ -1,4 +1,5 @@
 using Gum.Forms.Data;
+using Shouldly;
 using Xunit;
 
 namespace MonoGameGum.Tests.Forms.Data;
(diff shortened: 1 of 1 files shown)


MonoGameGum.Tests/Forms/Data/BinderHelpersTests.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        bool canWrite = BinderHelpers.CanWritePath(typeof(TestPathRoot), nameof(TestPathRoot.ReadonlyText));
+        Assert.False(canWrite);
+    }
+
+    [Fact]
+    public void CanWritePath_ReturnsTrue_ForWritableProperty()
+    {
+        bool canWrite = BinderHelpers.CanWritePath(typeof(TestPathRoot), nameof(TestPathRoot.Text));
+        Assert.True(canWrite);
```
vchelaru: Can we use Shouldly extensions like canWrite.ShouldBeTrue(); to match the style of other tests?

PR #2228, opened 2026-02-20
Document gum tool gum project files

### LINKED ISSUE #496: Screens and components deleted remain in gum_events.json
After I delete a screen or component, it remains in the gum_events.json file.

I don't think they should remain here.

EDIT: I think I just realized this is some sort of logging or something?  If so, suggest renaming to gum_events_log.json.  Also suggest creating a section in the doc that explains all the files, their purpose, and structure in Gum Tools.  Also suggest a .gitignore template or items uses should add to a .gitginore template

kaltinril:
Documentation about project files fixes #496


kaltinril:
NOTE: I've never edited the docs from github, so not sure if this is good.

PR #2233, opened 2026-02-21
Remove scaling for the color picker individual controls, instead, sca…

### LINKED ISSUE #468: UI Scaling Fallout compilation (Compilation ticket)

kaltinril:
…le the entire thing.

Fixes #468

### DIFF
--- a/Gum/Themes/AppScale.cs
+++ b/Gum/Themes/AppScale.cs
@@ -40,6 +40,9 @@ public sealed class AppScale : DependencyObject
     public double IconButton { get => (double)GetValue(IconButtonProperty); private set => SetValue(IconButtonProperty, value); }
     public double ToggleDisplayIcon { get => (double)GetValue(ToggleDisplayIconProperty); private set => SetValue(ToggleDisplayIconProperty, value); }
 
+    // Scale ratio (1.0 at default, useful for LayoutTransform on imported controls)
+    public double ScaleFactor { get => (double)GetValue(ScaleFactorProperty); private set => SetValue(ScaleFactorProperty, value); }
+
     public static readonly DependencyProperty BodyProperty = DP(nameof(Body), BaseFontSizeDefault * BodyScale);
     public static readonly DependencyProperty CaptionProperty = DP(nameof(Caption), BaseFontSizeDefault * CaptionScale);
     public static readonly DependencyProperty H1Property = DP(nameof(H1), BaseFontSizeDefault * H1Scale);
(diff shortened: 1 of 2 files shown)


vchelaru:
Looks beautiful!

PR #2264, opened 2026-02-24
1491 add right click lock

### LINKED ISSUE #1491: Add Lock Visual To context menu
**Current Behavior** : 

To lock an visual you need to use the Locked toggle on the variables window.

Image

**Desired behavior** : 

Add a context menu when right clicking on the added visual  and have an option that allow to toggle lock and unlock the visual from being selected.

example : 

Image

**Why the requested feature**: 

It would allow a more seamless workflow and make it faster and less menu seeking. 
Adding an lockpad to the object to show its status for clarity purposes would also be appreciated.

kaltinril:
fixes #1491

### DIFF
--- a/Gum/Plugins/InternalPlugins/VariableGrid/ElementSaveDisplayer.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/ElementSaveDisplayer.cs
@@ -107,17 +107,15 @@ public class ElementSaveDisplayer
     {
         var currentState = _selectedState.SelectedStateSave;
         bool isDefault = currentState == _selectedState.SelectedElement.DefaultState;
-        if (instanceSave?.DefinedByBase == true)
-        {
-            isDefault = false;
-        }
+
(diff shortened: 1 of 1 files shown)


Gum/Plugins/InternalPlugins/VariableGrid/SetVariableLogic.cs
```diff
@@ -45,6 +45,7 @@ public class SetVariableLogic : ISetVariableLogic
         {"BaseType",                                               VariableRefreshType.FullGridRefresh   },
         {"IsRenderTarget",                                         VariableRefreshType.FullGridRefresh   },
         {"TextOverflowVerticalMode",                               VariableRefreshType.FullGridRefresh   },
+        {"Locked",                                                 VariableRefreshType.FullGridRefresh   },
```
kaltinril: @vchelaru please make sure this isn't a bad change

PR #2275, opened 2026-02-25
2267 fix multli select delete

### LINKED ISSUE #2267: Multi-select behavior+component should ask to delete all
1. Select a component
2. CTRL+select a behavior
3. Press DEL

Observe - popup only asks about deleting behavior
Expected - popup should ask about deleting both, and clicking OK should delete both

kaltinril:
fixes #2267


kaltinril:
I think it's ready @vchelaru 
image

PR #2276, opened 2026-02-25
(DIscussion) Folder tag refactor

kaltinril:
I didn't like that folders in the treeview had no tag, so I added an enum to make checks more intelligent than "null"

If you like it, except it and merge it ONLY AFTER the other PR's

Setting as a draft so you can think about it


kaltinril:
per discussion i'm canceling this @vchelaru

PR #2280, opened 2026-02-26
Rename some winforms things to wpf naming conventions

kaltinril:
Some renames for winforms old controls that are now wpf


vchelaru:
Looks good, thanks!

PR #2383, opened 2026-03-17
Fix enum handling and UI refresh for TextureAddress

zaifyr:
Changing the TextureAddress property in the editor could result in the value being stored as a string ("Custom") instead of the enum TextureAddress.Custom. This caused two issues.

Runtime crash

```
InvalidCastException: Unable to cast object of type 'System.String'
to type 'Gum.Managers.TextureAddress'
```

and Property grid inconsistency

The coordinate fields (TextureLeft, TextureTop, TextureWidth, TextureHeight) were conditionally visible when:

`TextureAddress == Custom`

Because the state sometimes stored "Custom" as a string, the visibility rule failed and the fields would not appear until the element was reselected. This also effected the generated code but it was pretty hard to get that far.

I tired a couple older versions, all the way back to Nov. Not sure why I was running into this issue when it didn't seem widespread, but I'm pretty sure I was using the tool correctly.

Tried to get to the root of the issue, but I likely didn't find it. I settled for converting dropdown string values to the correct enum type before storing them in the state
in `HandleCustomSet` and a defensive conversion in `MainEditorTabPlugin.HandleVariableSetLate`
to convert any remaining string values to the correct enum before applying them to the runtime object.
 
 There probably could be a better way to handle this issue, but so far this fix is working for me.

### DIFF
--- a/Gum/Plugins/InternalPlugins/VariableGrid/StateReferencingInstanceMember.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/StateReferencingInstanceMember.cs
@@ -752,6 +752,26 @@ public class StateReferencingInstanceMember : InstanceMember
                     var rootVariable = ObjectFinder.Self.GetRootVariable(this.Name, elementSave);
                     variableType = rootVariable?.Type;
                 }
+
+                // Convert enum values coming from dropdown UI
+                if (newValue is string s && !string.IsNullOrEmpty(variableType))
+                {
+                    var enumType = Type.GetType($"Gum.Managers.{variableType}")
+                                   ?? Type.GetType(variableType);
+
+                    if (enumType != null && enumType.IsEnum)
+                    {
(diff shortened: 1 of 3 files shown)


vchelaru:
Thank you for reporting and researching a fix here. I did a little bit of research myself and found the root of the problem.

The issue comes from how the Texture Address dropdown behaves. By default you can see it is a normal dropdown:

image

Selecting a new value here results in the texture address mode being correctly assigned.

However, if you switch between the texture address modes a few times, the combo box becomes editable:

image

This can also happen if you deselect and re-select the sprite.

When this happens, the combo box reports its value as a string rather than an enum value because the user could type anything.

This is the core root of the problem.

The fix is to find why the combo box can become editable in the first place and prevent it from ever doing so, which will keep Gum from assigning a string value.

I'm investigating now and will follow up with my findings.

PR #2498, opened 2026-04-10
2492 add support for specifying ttf file path for font

### LINKED ISSUE #2492: Add support for specifying .ttf file path for font
Currently you have to have the font installed. There should be a way to do a font that is a file path instead of just a system font so that users don't have to install fonts to open gum projects. This could work with either bmfont or KS, but let's verify. If not, then we would require KS?

vchelaru:


vchelaru:
This is Phase A, just runtime no Gum tool so far.

PR #2540, opened 2026-04-19
Fix text wrap in SokolGumSample

wiredmatt:
Before
image

---
After
image


wiredmatt:
CI failed because I accidentally deleted my fork of Sokol.NET. Forked it again.

I'll probably trim that fork in the future, it pulls in a lot of stuff for the examples it contains, fetching the submodule fully can take a few minutes. It also includes dylibs in binary form, I'll see what to do about that later on... all future improvements

PR #2674, opened 2026-05-08
Fix DefaultChildContainer leaking onto component instances (#2672)

### LINKED ISSUE #2672: DefaultChildContainer ("Default Slot") variable leaks onto instances of a component
## Repro

1. Open a project containing a screen or component that has a `ComboBox` instance (any Forms component with `DefaultChildContainer` set works — `ComboBox`, `ListBox`, `ScrollViewer`, `Menu`, etc.).
2. Select the `ComboBox` component itself → variable grid shows **Default Slot** (the `DefaultChildContainer` variable). ✅ expected.
3. Select an *instance* of `ComboBox` inside another container → variable grid shows a **Default Child Container** variable. ❌ should not appear here.

Both rows refer to the same underlying variable (`DefaultChildContainer`); the component-level display uses the friendly name "Default Slot", while the instance-level display falls through to the raw variable name.

## Why it happens

In `Gum/Plugins/InternalPlugins/VariableGrid/ElementSaveDisplayer.cs`:

- Line 159: `shouldShowChildContainer = effectiveElementSave is ComponentSave && instanceSave == null` — correctly scopes the variable to the component itself.
- Lines 236–241: when `!shouldShowChildContainer`, the code removes `DefaultChildContainer` from the property list **and from `addedNames`**.
- Lines 268+: the main variable-iteration loop then walks `defaultState.Variables` (the ComboBox's default state, which has `DefaultChildContainer` set in `ComboBox.gucx`). Because `addedNames` no longer contains it, the dedup guard in `CreatePropertyData` (line 944, `shouldInclude &= !addedNames.Contains(defaultVariable.Name)`) lets it through, and the variable is re-added as a regular variable on the instance.

So the early removal block is fighting the later additive loop. The fix is to either keep the name in `addedNames` (so dedup blocks the re-add) or to filter `DefaultChildContainer` out inside the per-variable loop when `instanceSave != null`.

## Expected behavior

`DefaultChildContainer` (a.k.a. "Default Slot") should only appear when the component itself is selected, never on instances of that component.

vchelaru:
## Summary
- DefaultChildContainer ("Default Slot") was appearing in the variable grid for instances of any component that set it (ComboBox, ListBox, ScrollViewer, Menu, ...). It should only appear on the component itself.
- Root cause in `ElementSaveDisplayer.FillPropertyList`: when not showing the slot, the code removed `DefaultChildContainer` from `addedNames`, so the later per-variable loop re-added it from the component's default state. Switched to adding it to `addedNames` so the existing dedup guard in `CreatePropertyData` blocks the re-add.

## Test plan
- [x] New unit test `GetCategories_InstanceOfComponentWithDefaultChildContainer_DoesNotLeakDefaultChildContainerVariable` repros the bug (red without the fix, green with it).
- [x] Existing tests in `ElementSaveDisplayerFormsPropertiesTests` still pass.
- [x] Full `GumFull.sln` test suite passes.

Closes #2672

### DIFF
--- a/Gum/Plugins/InternalPlugins/VariableGrid/ElementSaveDisplayer.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/ElementSaveDisplayer.cs
@@ -232,12 +232,14 @@ public class ElementSaveDisplayer
         }
 
 
-        // now that variables have been added we can remove the default child container:
+        // DefaultChildContainer (a.k.a. "Default Slot") should only appear on the component
+        // itself, never on instances or screens. Add it to addedNames so the per-variable
+        // loop below dedups it instead of re-adding it from the component's default state.
         if(!shouldShowChildContainer)
         {
             string variableName = "DefaultChildContainer";
             propertyList.RemoveAll(item => item.OriginalName == variableName);
-            addedNames.Remove(variableName);
+            addedNames.Add(variableName);
         }
 
         #region Loop through all variables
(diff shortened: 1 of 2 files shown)


vchelaru:
Tested, it's working properly

Issue #2635, opened 2026-05-07
Add CircleRuntime support to RaylibGum via direct source link


vchelaru:
## Goal

`CircleRuntime` exists in MonoGameGum and SkiaGum but not in RaylibGum. Close that gap by file-linking the existing `MonoGameGum/GueDeriving/CircleRuntime.cs` into `RaylibGum.csproj`, adding any `#if RAYLIB` guards needed for Raylib-specific differences (primarily the `Color` type).

## Scope

- **Direct source link** — `<Compile Include="..\..\MonoGameGum\GueDeriving\CircleRuntime.cs" Link="GueDeriving\CircleRuntime.cs" />` in `Runtimes/RaylibGum/RaylibGum.csproj`. No duplicate file.
- **Namespace stays as-is.** The shared file already declares `Gum.GueDeriving` (after PR #2634 lands), and Raylib already uses that namespace. No namespace work needed.

> [!IMPORTANT]
> This issue should be picked up **after PR #2634 (runtime namespace unification) merges.** That PR moves all runtime classes into `Gum.GueDeriving`. Starting before it merges would mean reworking `#if` guards on the namespace declaration.

## Why this is small

- Raylib **already has the underlying renderable**: `Runtimes/RaylibGum/Renderables/LineCircle.cs`. The Raylib `LineCircle` exposes the same `Radius` and `Color` surface that MonoGame's does, so the runtime-level wrapper should compile against it with minimal `#if` guards.
- The same convergence pattern was already done for `SpriteRuntime`, `TextRuntime`, `ContainerRuntime`, `ColoredRectangleRuntime`, and `NineSliceRuntime`. `CircleRuntime` is a smaller file (~137 lines) than any of those, so the playbook is well-rehearsed.

## Files to touch

1. `Runtimes/RaylibGum/RaylibGum.csproj` — add the `<Compile Include>` line.
2. `MonoGameGum/GueDeriving/CircleRuntime.cs` — add `#if RAYLIB` guards as needed.

## Known platform difference

The `Color` property currently exposes `Microsoft.Xna.Framework.Color`. Raylib uses `Raylib_cs.Color`. Follow the same pattern used in `SpriteRuntime.cs` / `TextRuntime.cs`:

```csharp
#if RAYLIB
using Color = Raylib_cs.Color;
#elif SKIA
using Color = SkiaSharp.SKColor;
#else
using Color = Microsoft.Xna.Framework.Color;
#endif
```

[...]

PR #2679, opened 2026-05-09
2635 raylib circle runtime

atiaen:
# Summary 

Implementation for #2635 

  - Replaces the `LineCircle `stub in RaylibGum with a working implementation, enabling `CircleRuntime `to function on Raylib
  - Updates `CircleRuntime `to handle Raylib's value-type `Color `struct correctly in the `Alpha, Red, Green, and Blue` setters
  - Updates `CircleRuntime `to resolve `LineCircle `from `Gum.Renderables `instead of `RenderingLibrary.Math.Geometry` when
  targeting Raylib

`  Runtimes/RaylibGum/Renderables/LineCircle.cs` was previously a stub that threw `NotImplementedException`. It now:

  - Inherits from `InvisibleRenderable `which provides position, size, visibility, and parent/child wiring via `RenderableBase`
  - Renders using Raylib's `DrawCircleLines`, with the centre calculated from the absolute top-left position
  - Supports `CircleOrigin.Center` and `CircleOrigin.TopLeft` — when `TopLeft`, `Radius `is added to both axes to derive the true
  centre
  - Defaults `Color` to `Color.White`

  **Test plan**

  - `CircleRuntimeTests `contains tests for similar to `ContainerRuntimeTests`


vchelaru:
@atiaen — pushed several follow-up commits on top of yours. Wanted to walk through the reasoning so nothing feels arbitrary.

### `remove_gum.sh` / `setup_gum_linux.sh` mode changes
These showed up in the diff as 755 → 644 (likely a side-effect of editing the branch on Windows). Restored the executable bit so the diff stays scoped to the actual feature.

### `CircleRuntime.cs` cleanup
- Dropped the stray `using Gum.RenderingLibrary;` from the `#else` branch — that namespace contains `Blend`/`BlendExtensions`, not `LineCircle`, so it wasn't pulling its weight.
- Aligned the file structure with `ColoredRectangleRuntime.cs`: added the `#if MONOGAME || FNA || KNI \n #define XNALIKE` shim at the top (defensive — keeps the file safe if it ever gets linked into a project that forgot to set the symbol), and aliased the contained renderable as `ContainedCircleType` per platform.
- Fixed the broken indentation on the `#if XNALIKE` `Color` getter (`         get` had 9 leading spaces).
- Added `NotifyPropertyChanged()` to the `Alpha`/`Red`/`Green`/`Blue` setters. The original file was missing these — pre-existing inconsistency, fixed boyscout-style since the file was already being restructured. `ColoredRectangleRuntime` calls them on every channel setter, so this brings circle in line.

### `Runtimes/RaylibGum/Renderables/LineCircle.cs`
- Reworked to match `NineSlice`/`Sprite` style: `Color` is now a `{ get; set; } = Color.White;` property and there are `Alpha/Red/Green/Blue` helpers using the `Color = new Color(R, G, B, A)` reconstruction pattern. Consistency with the other Raylib renderables.
- **Behavior fix**: `Render` was anchoring the centre at `(AbsoluteLeft + Radius, AbsoluteTop + Radius)` in `CircleOrigin.TopLeft` mode. That silently assumes `Width == Height == 2 * Radius`. It holds when callers only set `Radius` through `CircleRuntime` (whose setter syncs Width/Height), but breaks the moment those diverge — e.g. `WidthUnits = RelativeToParent` with `Width = 0`, animations driving Width independently, or any caller setting Width after Radius. Switched to deriving the centre from the actual extents (`AbsoluteLeft + Width * 0.5f`, etc.) so the circle stays centred regardless of how Width/Height were set.

### Re: my earlier comment about `WithAlpha` and value types
I'd framed the per-channel `#if RAYLIB` branches as a value-type-struct issue. That was wrong. The actual reason is that `ToolsUtilitiesStandard.Helpers.ColorExtensions.WithAlpha` is declared as an extension on `System.Drawing.Color` (top of `ToolsUtilities/Helpers/ColorExtensions.cs`). Raylib's `LineCircle.Color` is `Raylib_cs.Color` — a different type — so the extension method isn't in scope on the Raylib branch. Your inline approach is the right call; adding `Raylib_cs.Color` overloads would force `ToolsUtilities` to take a Raylib dependency.

### Tests
Expanded `CircleRuntimeTests` from 4 tests to 9: added `Color_ShouldDefaultToWhite`, `Color_ShouldRoundTrip`, and `Red/Green/Blue_ShouldRoundTrip`. Renamed `cut` → `sut` to match `ContainerRuntimeTests` and added the trailing newline. The original 4 covered the runtime existing; these cover the Raylib-specific behavior the PR introduces (per-channel writes against the value-type Color struct). Full RaylibGum.Tests suite: 103/103 passing.

### Sample (`Samples/raylib/Program.cs`)
This was scope creep — the user wanted a place to actually demonstrate `CircleRuntime` and felt the existing sample setup was overdue for structure anyway. Two screens that swap on SPACE: a "Raw visuals" screen (Sprites row, Shapes row, Centering demo) and the existing Forms controls. The Centering row places a `CircleRuntime` inside a 150×150 `ColoredRectangleRuntime` with `RelativeToParent` units and `Radius = 30` — this is what visually confirmed the `LineCircle.Render` fix above (without the fix, the orange circle anchors to the top-left of the blue box; with it, it sits in the middle). All container helpers explicitly set `Width = 0, Height = 0` because `ContainerRuntime` defaults to 150×150 and with `RelativeToChildren` units those values become *padding added to* the children's bounding box.

Thanks again for the implementation — the rendering work itself was straightforward and correct, the follow-up was mostly polish and one real centering bug that only surfaces when Width/Height diverge from `2 * Radius`.

PR #2760, opened 2026-05-14
Spike #2758: renderable-swap mechanism for a fill-capable CircleRuntime

### LINKED ISSUE #2758: Spike: renderable-swap mechanism for a fill-capable CircleRuntime (MonoGame + Apos.Shapes)
## Background

Part of the runtime type-collapse effort — see `.claude/designs/runtime-unification/TypeCollapse.md`.

Long-term goal: `CircleRuntime` becomes a fully-featured circle on every backend (filled **or** outline), driven by the decided API — `Color? FillColor`, `Color? StrokeColor`, `float StrokeWidth` (never `IsFilled`). On MonoGame, the filled-circle capability comes from **Apos.Shapes**, which is an *optional* package — core `MonoGameGum` does not and should not reference it.

## Why a spike, and why first

The API shape is already decided, so it isn't the risk. The genuine unknown is the **renderable-acquisition + swap mechanism** on MonoGame:

> How does a single `CircleRuntime` type acquire a fill-capable renderable from an optional package, without core `MonoGameGum` referencing that package?

Doing the easy backend first (Raylib has native fill; Skia always has SkiaSharp) would be a quick win — but if the MonoGame + Apos.Shapes mechanism then turns out awkward or impossible to do cleanly, we'd be painted into a corner with a shipped API/structure that doesn't generalize. So we de-risk the load-bearing unknown **before** committing any backend.

The spike is inherently MonoGame + Apos.Shapes work — that's the only place the optional-dependency boundary exists.

## The key question to resolve

`TypeCollapse.md` currently recommends **composite** (the runtime owns both an outline and a fill renderable and toggles visibility) for the simple Circle/Rectangle collapse, reserving the **capability-factory** approach for the later rounded-rect fold.

But composite implies core `CircleRuntime` constructs the fill renderable itself → core references Apos.Shapes → not allowed. So the spike must determine:

- Does **composite** survive the optional-dependency boundary for `Circle` at all?
- Or does `Circle` need the **capability-factory / registration** approach *now* — earlier than `TypeCollapse.md` assumes?

If it's the latter, that finding rewrites the doc's sequencing assumption.

## Scope — keep it tight

**In scope:**
- A working mechanism for `CircleRuntime` to render filled (via Apos.Shapes when present) or outline, behind the decided `FillColor` / `StrokeColor` / `StrokeWidth` API, across the optional-dependency boundary.
- Graceful behavior when Apos.Shapes is not referenced (outline-only, no crash).

**Explicitly NOT in scope:**
- Folding `ColoredCircleRuntime` into `CircleRuntime` (the actual collapse).
- The Roslyn analyzer / `[Obsolete]` shim migration path.
- Raylib or Skia.
- API polish, docs site, tool-side changes.
- `RectangleRuntime` / `PolygonRuntime`.

Throwaway-quality code is acceptable — **the deliverable is a decision (composite vs capability-factory), backed by working proof-of-concept code**, plus an update to `TypeCollapse.md` recording the outcome.

## After the spike

Once the mechanism is proven, the Raylib-vs-MonoGame ordering for the actual collapse becomes low-stakes (the hard part is solved). The spike already lives on MonoGame, so hardening continues there — or Raylib gets done as a quick win, now confident it generalizes. Skia convergence and the #2757 restructure also get sequenced off this finding.

## Related

- `.claude/designs/runtime-unification/TypeCollapse.md` — the collapse design
- #2757 — Skia shape-runtime unification (currently mis-scoped; to be restructured based on this spike's outcome)

vchelaru:
## Spike #2758 — outcome

**Decision: capability-factory / registration, not composite, for the `Circle` collapse.**

The spike's load-bearing question was: *how does a single `CircleRuntime` (in core `MonoGameGum`) acquire a fill-capable renderable from the optional Apos.Shapes package, without core referencing that package?*

**Composite is dead on arrival for `Circle`.** Composite means core `CircleRuntime` owns both renderables and toggles between them — but the fill renderable (`MonoGameAndGum.Renderables.Circle`) lives in `MonoGameGumShapes`, and the dependency arrow is `MonoGameGumShapes → MonoGameGum` only. Core can't `new Circle()` — can't even name the type. This contradicts `TypeCollapse.md`'s "composite first" recommendation *for `Circle` specifically*; `Rectangle` is unaffected (both its renderables ship in core).

**The good news:** the capability-factory is *one more* of an extension pattern the codebase already uses for this exact optional-package boundary — `ElementSaveExtensions.RegisterGueInstantiation`, `StandardElementsManager.CustomGetDefaultState`, `CustomSetPropertyOnRenderable.AdditionalPropertyOnRenderable`. All core-defined hooks, all filled by `MonoGameGumShapes` from the same `[ModuleInitializer]`. And `GraphicalUiElement.SetContainedObject` already does the renderable rewire. Low risk.

## Proof-of-concept (throwaway-quality, FillColor-only per spike scope)

- **Core `MonoGameGum`** (`#if XNALIKE`): `IFilledShapeRenderable` capability contract + `ShapeRenderableRegistry` static extension point.
- **`CircleRuntime.FillColor`** (`Color?`) drives the swap via `SetContainedObject`: builds a fill-capable renderable from the registry when present; stays on the outline `LineCircle` when absent — graceful degradation, no crash.
- **`MonoGameGumShapes`**: `RenderableShapeBase` implements `IFilledShapeRenderable`; `AposShapeRuntime.RegisterRuntimeTypes()` registers `() => new Circle()`.

## Tests — both sides of the boundary

- `MonoGameGum.Shapes.Tests/CircleRuntimeFillColorTests` — the swap works (renderable becomes the Apos `Circle`, color forwarded, swaps back to `LineCircle` on null).
- `MonoGameGum.Tests/Runtimes/CircleRuntimeTests` — graceful degradation when `MonoGameGumShapes` is *not* referenced (no crash, stays outline).

Full `MonoGameGum.Shapes.Tests` (21) and `MonoGameGum.Tests` Runtimes (649) green. `KniGum` / `KniGumShapes` / `RaylibGum` build clean. `FnaGum` not verified — the FNA submodule isn't checked out in this worktree — but it follows the same glob + `XNALIKE` path with no shapes package, landing on graceful degradation.

## Findings for the *actual* collapse (out of scope here; recorded in `TypeCollapse.md`)

1. Registration is assembly-load-triggered, and `CircleRuntime` lives in a different assembly than the registrant — using `CircleRuntime` alone doesn't load `MonoGameGumShapes`. The collapse should make the load-order contract explicit.
2. The `internal OnPreRender` hook can't be wired from core — `StrokeWidth` with `ScreenPixel` units will need the factory signature to become `Func<GraphicalUiElement, IFilledShapeRenderable>`.
3. `CircleRuntime`'s existing `Color`/`Radius`/`Alpha` read through a `LineCircle` cast that returns null post-swap — routing those is the bulk of the real collapse work.

`TypeCollapse.md` lives under the gitignored `.claude/designs/` tree, so its update isn't in this diff — it's been updated in the local checkout with the full outcome writeup.

Closes #2758

### DIFF
--- a/MonoGameGum.Tests/Runtimes/CircleRuntimeTests.cs
+++ b/MonoGameGum.Tests/Runtimes/CircleRuntimeTests.cs
@@ -1,0 +1,24 @@
+using Gum.GueDeriving;
+using Microsoft.Xna.Framework;
+using RenderingLibrary.Math.Geometry;
+using Shouldly;
+using Xunit;
+
+namespace MonoGameGum.Tests.Runtimes;
+
+public class CircleRuntimeTests : BaseTestClass
+{
+    // Spike (#2758): when the optional MonoGameGumShapes package is NOT referenced (as is the
+    // case for this test project), the ShapeRenderableRegistry factory is never populated.
+    // Setting FillColor must degrade gracefully — no crash, and the runtime stays on its
+    // outline LineCircle renderable.
+    [Fact]
+    public void FillColor_WhenSetWithoutShapesPackage_DoesNotThrowAndStaysOutline()
+    {
+        CircleRuntime sut = new();
+
+        Should.NotThrow(() => sut.FillColor = Color.Red);
+
+        sut.RenderableComponent.ShouldBeOfType<LineCircle>();
+    }
+}
--- a/MonoGameGum/GueDeriving/CircleRuntime.cs
+++ b/MonoGameGum/GueDeriving/CircleRuntime.cs
@@ -114,6 +114,78 @@ public class CircleRuntime : GraphicalUiElement
 #endif
     }
 
+#if XNALIKE
+    Color? _fillColor;
+
+    /// <summary>
+    /// When non-null, the circle renders as a solid fill of this color. When null (the default),
+    /// the circle renders as an outline via its <see cref="LineCircle"/> renderable.
+    /// </summary>
+    /// <remarks>
+    /// Filled rendering requires the optional MonoGameGumShapes (Apos.Shapes) package, which
+    /// registers a renderable factory with <see cref="ShapeRenderableRegistry"/>. When that
+    /// package is not referenced, setting <see cref="FillColor"/> is a graceful no-op for the
+    /// visual: the runtime stays on its outline renderable rather than crashing.
+    ///
+    /// Spike (#2758): proof-of-concept for the renderable-swap mechanism only. It does not yet
+    /// route the existing Color/Radius/Alpha properties through a swapped-in fill renderable —
+    /// that wiring is part of the actual ColoredCircle/Circle collapse, which is out of scope
+    /// for the spike.
+    /// </remarks>
+    public Color? FillColor
+    {
+        get => _fillColor;
+        set
+        {
+            _fillColor = value;
+            UpdateRenderableFromFillColor();
+            NotifyPropertyChanged();
+        }
+    }
+
+    /// <summary>
+    /// Swaps the contained renderable to match the current <see cref="FillColor"/>: a
+    /// fill-capable renderable (from <see cref="ShapeRenderableRegistry"/>) when FillColor is
+    /// set, or the outline <see cref="LineCircle"/> when it is null. This is the core of the
+    /// renderable-swap mechanism the spike exists to prove.
+    /// </summary>
+    void UpdateRenderableFromFillColor()
+    {
+        if (_fillColor.HasValue)
+        {
+            if (RenderableComponent is IFilledShapeRenderable existingFill)
+            {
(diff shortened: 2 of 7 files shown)


vchelaru:
Closing unmerged. The spike succeeded — its deliverable was the validated *decision*, recorded in `TypeCollapse.md` ("Spike Outcome") and `RuntimeNorthStar.md`. The POC proved the mechanism builds and works, but `ShapeRenderableRegistry` / `IFilledShapeRenderable` are shape-specific scaffolding we've since agreed to generalize into a `RenderableRegistry` — merging them would ship known-throwaway public API into the `Gum.MonoGame` package. The branch stays as a reference artifact. Full implementation tracked in #2761.

PR #2767, opened 2026-05-15
Phase 2 of #2761: collapse ColoredCircle into CircleRuntime via RenderableRegistry

### LINKED ISSUE #2761: RenderableRegistry: generalize renderable swapping and collapse ColoredCircleRuntime into CircleRuntime
## Background

Follows the spike in #2758 (PR #2760, closed unmerged — the spike's deliverable was the validated *decision*, not shippable code). Design docs: `.claude/designs/runtime-unification/TypeCollapse.md` and `RuntimeNorthStar.md`.

The spike proved the renderable-acquisition + swap mechanism for collapsing `ColoredCircleRuntime` into `CircleRuntime` on MonoGame, across the optional Apos.Shapes (`MonoGameGumShapes`) dependency boundary that core `MonoGameGum` must not cross.

## Decisions carried in from the spike + follow-up discussion

1. **Capability-factory / registration, not composite.** Composite is dead on arrival for `Circle`: it requires core `CircleRuntime` to construct the Apos.Shapes `Circle` renderable, which lives in the optional package core cannot reference or even name. (Composite still applies to `Rectangle` — both its renderables ship in core.)

2. **The registry generalizes beyond shapes.** The spike used a shape-specific `ShapeRenderableRegistry`. The real mechanism is a general `RenderableRegistry` — the swap pattern is valuable for any renderable. The motivating future case: swapping Gum's native `Text` for a FontStashSharp-backed renderable.

3. **Two registry semantics are needed, because the Circle and Text cases differ:**
   - **Factory-or-nothing** (Circle): the capability is *boundary-forced* — core can't construct it. Factory present → use it; absent → graceful degradation.
   - **Default-plus-override** (Text): the capability is *not* boundary-forced — `Text` ships in core and works. Core registers a default; a consumer or optional package may *replace* it.
   `RenderableRegistry` must support both.

4. **Capability interfaces over role interfaces.** `IFilledShapeRenderable` (from the spike) is a small, composable *capability* contract — prefer that style. A wide `ITextRenderable` "whole renderable" *role* interface is brittle: every member becomes a third-party stability commitment. Let interface width be *forced* by a real consumer, not designed speculatively.

5. **Demand-driven, one renderable at a time.** This issue does `Circle`. Other renderables (Text/FontStashSharp, etc.) get their interfaces when a real second implementor pins the surface — tracked separately as a north-star arc.

### Phase 1 — `RenderableRegistry`
- Generalize `ShapeRenderableRegistry` → `RenderableRegistry` in core `MonoGameGum`.
- Support both **factory** (absent-by-default) and **default-override** (present-by-default) registration semantics.
- Keep the core-defines-the-contract / optional-package-implements pattern: mirrors `ElementSaveExtensions.RegisterGueInstantiation`, `StandardElementsManager.CustomGetDefaultState`, `CustomSetPropertyOnRenderable.AdditionalPropertyOnRenderable`.

### Phase 2 — full `Circle` collapse via the registry
- `CircleRuntime` exposes the decided API: `Color? FillColor`, `Color? StrokeColor`, `float StrokeWidth` (+ units). No `IsFilled` on the runtime — see `RuntimeNorthStar.md` → "Nullable fill/stroke at the runtime layer".
- `CircleRuntime` swaps its contained renderable through the registry: a fill-capable Apos renderable when the API requires it, the outline `LineCircle` otherwise. Graceful degradation when `MonoGameGumShapes` is absent.
- **Route the existing `Color` / `Radius` / `Alpha` / `Red` / `Green` / `Blue` properties through whichever renderable is currently contained.** Today they read through a `LineCircle` cast that returns null post-swap — this is the bulk of the real collapse work.
- **Fix the `OnPreRender` hook.** `AposShapeRuntime.SetContainedShape` wires the renderable's `internal OnPreRender` so the runtime's `PreRender` fires (needed for `ScreenPixel` `StrokeWidth`). Core can't see `internal`. Likely fix: the factory signature becomes `Func<GraphicalUiElement, IFilledShapeRenderable>` (or similar) so the optional assembly does the assembly-internal wiring core can't.
- **Define the registration load-order contract.** `CircleRuntime` lives in a different assembly than the registrant, so using `CircleRuntime` alone doesn't load `MonoGameGumShapes`. Make the contract explicit rather than relying on incidental assembly touch (same caveat already noted on `AposShapeRuntime.RegisterRuntimeTypes`).
- TDD: swap behavior, graceful degradation, and legacy-property routing — both sides of the boundary (`MonoGameGum.Shapes.Tests` and `MonoGameGum.Tests`).

## Spike findings to carry forward
(from #2758 / `TypeCollapse.md` → "Spike Outcome")

1. Assembly-load-triggered registration vs. cross-assembly `CircleRuntime` — needs an explicit load-order contract (Phase 2).
2. `internal OnPreRender` hook core can't wire — factory signature change (Phase 2).
3. Legacy `Color`/`Radius` properties NRE through the `LineCircle` cast post-swap (Phase 2).
4. Spike was gated `#if XNALIKE` and MonoGame-tested only. KNI/`KniGumShapes` build clean; FNA was unverified (FNA submodule not checked out in the spike worktree). Confirm the FNA build; decide whether the gate stays `XNALIKE` or narrows.
5. `RenderableRegistry` as a static — consistent with the sibling extension points, but `code-style.md` prefers interfaces/DI. Deliberate call needed (Phase 1).

## Open questions
- **Global vs per-instance swap.** A single static registry makes swapping all-or-nothing per type ("FontStashSharp everywhere" works, "FontStashSharp for this label, native for that one" doesn't). Per-instance override is a harder design — decide deliberately.
- **Static registry vs DI** (finding #5).

## Out of scope — separate follow-ups
- **Migration path** — `[Obsolete]` `ColoredCircleRuntime` shim + Roslyn `GUM###` analyzer / code-fix. `TypeCollapse.md` → "Migration Approach" describes it.
- **Per-renderable interfaces / FontStashSharp-style text swapping** — the broader north-star arc. Demand-driven, later.
- **Raylib / Skia `Circle` collapse** — sequenced after the MonoGame mechanism lands (low-stakes once proven).
- **`RectangleRuntime` / `RoundedRectangleRuntime` collapses** — their own `RuntimeNorthStar.md` workstream #3 items.

## Related
- #2758 — the spike (closed)
- #2757 — Skia shape-runtime unification (to be re-scoped against this)
- `.claude/designs/runtime-unification/TypeCollapse.md`
- `.claude/designs/runtime-unification/RuntimeNorthStar.md`

vchelaru:
Extends Phase 1 (#2766 — general-purpose `RenderableRegistry`) with the actual Circle collapse on MonoGame.

## What

`CircleRuntime` (XNALIKE) gains the locked-in nullable fill/stroke API and swaps its contained renderable through `RenderableRegistry` to acquire a fill-capable Apos.Shapes renderable across the optional-dependency boundary.

- `Color? FillColor` — non-null → swap to `IFilledShapeRenderable` with `IsFilled=true`.
- `Color? StrokeColor` — non-null & no FillColor → `IFilledShapeRenderable` with `IsFilled=false`; falls back to writing the color onto the outline `LineCircle` when no factory is registered.
- `float StrokeWidth` + `DimensionUnitType StrokeWidthUnits` — held on the runtime; pushed to the fill renderable in `PreRender` with `AposShapeRuntime`-style ScreenPixel scaling.
- Legacy `Color` / `Red` / `Green` / `Blue` / `Alpha` are now `[Obsolete]` pointing at `FillColor` / `StrokeColor`, but stay functional via routing through whichever renderable is currently contained — this is the bulk of the "real collapse" work flagged in the spike findings.

`ColoredCircleRuntime` is **untouched** in this phase. Its `[Obsolete]` shim + GUM### Roslyn analyzer migration path is a separate follow-up; Raylib/Skia Circle collapse is also separate.

## Mechanism changes vs. spike #2758

| Spike finding | Phase 2 resolution |
|---|---|
| Spike's one-off `ShapeRenderableRegistry` | Migrated to general-purpose `RenderableRegistry` from Phase 1. |
| Core can't wire the renderable's `internal OnPreRender` | `RenderableRegistry` gains context-bearing factory overloads (`Func<GraphicalUiElement, T>` + `Create<T>(context)`). `MonoGameGumShapes` registers a factory that builds the `Circle` *and* wires `OnPreRender` to the caller GUE's `PreRender` in one shot — the assembly-internal access core can't do. |
| Legacy props NRE through stale `LineCircle` cast post-swap | `Color`/`Red`/`Green`/`Blue`/`Alpha`/`Radius` now check `RenderableComponent is IFilledShapeRenderable` first; `LineCircle` is the fallback. |
| Load-order — assembly-load-triggered registration doesn't survive `RenderableRegistry.Reset()` | `AposShapeRuntime.RegisterRuntimeTypes` registers the factory **outside** the `_registered` guard. `GumService.Initialize`'s reflection scan already re-invokes the public static method each cycle, so the factory is restored on every Initialize after Uninitialize. |
| `IFilledShapeRenderable` surface | Added `StrokeWidth` to `IsFilled` + `Color` so the per-frame ScreenPixel scaling can push through the capability without core seeing the Apos type. |

## Tests

TDD throughout; both sides of the optional-dependency boundary.

- **`MonoGameGum.Tests`** (no `MonoGameGumShapes` reference) — graceful degradation when no factory; legacy-property routing on `LineCircle`; 12 new `RenderableRegistry` tests covering both factory buckets and the four-tier resolution order. **1469 passed**.
- **`MonoGameGum.Shapes.Tests`** — full swap matrix (`FillColor` only / `StrokeColor` only / both → fill wins / both null → back to `LineCircle`), legacy `Color` routing through `Circle` post-swap, `StrokeWidth` propagation via the wired `OnPreRender` hook, and the load-order recovery test (`Reset` + recall `RegisterRuntimeTypes` → factory recovered, swap still works). **26 passed**.

## Decisions

- **`[Obsolete]` only under `#if XNALIKE`.** Raylib/Skia don't have `FillColor`/`StrokeColor` yet, so a deprecation warning pointing migrators at non-existent members would mislead. When those backends collapse, the gate widens.
- **`StrokeColor` graceful degradation** writes the stroke color onto `LineCircle.Color` — same slot the legacy `Color` setter uses. Users without the shapes package still see the intended color, just without the stroke-width / dash machinery.
- **`IFilledShapeRenderable` added to `GumCoreShared.projitems`.** Capability interface, not a generated runtime, so the `MonoGameGum/GueDeriving/*Runtime.cs` exclusion doesn't apply. If FRB1 builds get noisy this can be reverted with no functional impact.

## Out of scope (separate follow-ups)

- `[Obsolete]` shim of `ColoredCircleRuntime` and the GUM### Roslyn analyzer migration path.
- Raylib / Skia Circle collapse.
- Workstream #4 (fill + stroke in one shape).

Closes #2761


vchelaru:
Follow-up issue for the Rectangle collapse (covering both `ColoredRectangleRuntime` and `RoundedRectangleRuntime` under one role interface): #2768. Uses the construct-time-binding pattern this PR introduces.

PR #2845, opened 2026-05-19
Walk categorized instance states in GetVariableListRecursive

vchelaru:
## Summary

Stacked on top of #2844. When an instance has a categorized state set in the current state (e.g. `TextInstance.TextCategoryState = "H1"`), and that state on the instance's `BaseType` has its own `VariableReferences`, the lookup should prefer it over the containing element's `BaseType` chain.

**Before:** `GetVariableListRecursive`'s two branches (`shouldGoToBaseType` and `shouldGoToInstanceComponent`) were mutually exclusive. For `DefinedByBase=true` instances, `shouldGoToBaseType` always won, so the categorized-state walk never ran. Setting H1 on `UpgradeButton.TextInstance` had no visible effect on its references — the inspector kept showing Button's `TextInstance.VariableReferences` and the editor styling didn't shift.

**After:** A categorized-state walk runs first (mirroring the precedence `GetValueRecursive` already applies to scalar Variables at `StateSaveExtensionMethods.cs:103-188`). If a state variable like `TextCategoryState = "H1"` resolves to a matching state on the instance's type and that state has the list, it wins. The existing base-type / instance-type-default fall-through is unchanged, so no regression for instances without a categorized state set.

Guiding principle (per discussion): **most specific wins** — a categorized state on the instance is more specific than the containing element's BaseType list.

## Stacking

Targets `fix/variable-references-inheritance-display` (#2844). The fix only surfaces in the UI if #2844's `GetValueRecursive` change is also present, because the inspector reaches `GetVariableListRecursive` through `GetValueRecursive` on the default state. Once #2844 lands, this PR can rebase to `main` automatically.

## Test plan
- [x] New test `GetVariableListRecursive_OnInstance_ShouldPreferCategorizedStateOverBaseTypeList` — pins the H1 case.
- [x] New regression `GetVariableListRecursive_OnInstance_NoCategorizedStateSet_ShouldFallBackToBaseTypeList` — no categorized state set, base type walk still wins.
- [x] New regression `GetVariableListRecursive_OnInstance_CategorizedStateWithoutList_ShouldFallBackToBaseTypeList` — categorized state set but state has no list, base type walk still wins.
- [x] Full `GumToolUnitTests` suite passes (601 passed, 5 pre-existing skips, no regressions).
- [ ] Manual verification (UI, with #2844 merged): open `UpgradeButton.gucx`, select `TextInstance`. The `VariableReferences` row should reflect Label's H1 state (Font/FontSize/IsBold/IsItalic from Styles.H1.*) when `TextCategoryState = "H1"`. Editor preview should pick up H1's font/styling for properties not locally overridden in UpgradeButton.

### DIFF
--- a/Gum/DataTypes/StateSaveExtensionMethods.cs
+++ b/Gum/DataTypes/StateSaveExtensionMethods.cs
@@ -392,6 +392,50 @@ public static class StateSaveExtensionMethods
         {
             ElementSave elementContainingState = stateSave.ParentContainer;
 
+            // Most-specific-wins: if the variable is on an instance and the
+            // instance has a categorized state set in this state (e.g.
+            // TextInstance.TextCategoryState = "H1"), and that state on the
+            // instance's BaseType has the list, prefer it over the containing
+            // element's BaseType chain. This mirrors the precedence
+            // GetValueRecursive already applies for scalar Variables and lets
+            // VariableReferences cascade through categorized states.
+            if (elementContainingState != null && variableName.Contains('.'))
+            {
+                var sourceObjectName = VariableSave.GetSourceObject(variableName);
+                var instanceForStateWalk = elementContainingState.Instances
+                    .FirstOrDefault(item => item.Name == sourceObjectName);
+                if (instanceForStateWalk != null)
+                {
+                    var instanceType = ObjectFinder.Self.GetElementSave(instanceForStateWalk);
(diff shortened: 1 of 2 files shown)


vchelaru:
Folding into #2844 to keep everything in one branch and avoid squash-merge conflicts.

PR #3154, opened 2026-06-14
#566: Defer invalid variable-reference dialog so the variable grid isn't left showing a stale instance

### LINKED ISSUE #566: Variable References - Changing elements selected - Doesn't update display
So a bit of a small thing but if I enter an invalid variable reference, then I right-click on another element, the popup appears commenting out the variable reference, but then the element changes afterwards to the other element, leaving the the variable data displayed from the first instance.

this is a super super minor thing

https://github.com/user-attachments/assets/c8f3303c-a606-40f1-9f1d-6850c90be43f

vchelaru:
Closes #566

## Problem

When an incomplete/invalid variable reference (e.g. `X = ` with no right side) is typed into an instance's `VariableReferences` and the user then **selects another instance**, the edit is committed by focus loss. The validation correctly fails and comments the line out (`//X = `), but the failure popup appears **mid-commit** and the variable grid is left displaying the just-commented data on the *newly selected* instance. Selecting around afterward self-heals it, so the underlying data was always correct — this was a display/refresh-ordering glitch.

## Root cause

`VariableReferenceLogic.DoVariableReferenceReaction` showed the failure dialog **synchronously** in the middle of the commit. A modal dialog pumps the WPF message loop, so the pending selection-change cascade interleaves with the in-flight commit, and the grid ends up painted for the previously-selected instance.

## Fix

- Comment the invalid lines **synchronously** (data model stays consistent regardless of timing).
- **Defer** the failure dialog via `IDispatcher.Post` (background priority). This lets the commit and the pending selection change finish first; the deferred action then shows the dialog and re-refreshes the grid (`RefreshVariables(force: true)`) against the now-current selection.

The deferral path only runs when a reference is invalid (the common, valid path is untouched).

## Tests

- Added `DoVariableReferenceReaction_InvalidLine_CommentsImmediatelyButDefersFailureDialog`, which asserts the line is commented synchronously, the dialog is posted (not shown inline), and the deferred action both shows the dialog and refreshes the grid. Verified red before the fix (dialog shown inline → `Post` never called) and green after.
- Full `GumToolUnitTests` suite: 874 passed, 5 pre-existing skips, 0 failures.

## Note on verification

This is a WPF message-pump / refresh-ordering bug; the unit test proves the deferral mechanism, but the end-to-end timing behavior is best confirmed in the running tool. Worth a quick manual check of the original repro steps before merge.

### DIFF
--- a/Gum/Plugins/InternalPlugins/VariableGrid/VariableReferenceLogic.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/VariableReferenceLogic.cs
@@ -32,6 +32,7 @@ public class VariableReferenceLogic : IVariableReferenceLogic
     private readonly IDialogService _dialogService;
     private readonly IFileCommands _fileCommands;
     private readonly ICompositeMemberRegistry _compositeMemberRegistry;
+    private readonly IDispatcher _dispatcher;
 
     #endregion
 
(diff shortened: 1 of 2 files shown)


vchelaru:
Manually verified in the running tool: the original repro steps (type an incomplete reference like `X = `, then select another instance) no longer leave the commented `//X = ` displayed on the newly-selected instance.

PR #3169, opened 2026-06-15
#3167: Document ListBox keyboard/gamepad item-focus model

### LINKED ISSUE #3167: Research: directly give a ListBoxItem keyboard focus without 'entering' navigation
## Summary

Research whether a `ListBox` can hand keyboard focus directly to an individual `ListBoxItem`, so the user does **not** have to first focus the ListBox at the "top level" and then press a key to "enter" item-level navigation.

If it is possible, document it. If it is not possible, fix it and then document it.

## Background

`ListBox` (MonoGameGum/Forms/Controls/ListBox.cs) implements a two-level keyboard/gamepad focus model:

- The ListBox can be focused at the **top level** (`IsFocused`), and
- focus can then move down into the **individual items** (`DoListItemsHaveFocus`).

Relevant existing API surface that looks directly applicable:

- `DoListItemsHaveFocus` (public, settable) — its setter already forces `IsFocused = true` when set to `true`, and pushes `IsFocused` down onto `ListBoxItemsInternal[SelectedIndex]`.
- `CanListItemsLoseFocus` — when set to `false`, the setter forces `DoListItemsHaveFocus = true`, i.e. items always hold focus. Intended for the case where the ListBox is the only focusable thing on the screen.

This suggests the feature may already exist (set `DoListItemsHaveFocus = true`, and/or `CanListItemsLoseFocus = false`). The research should confirm the supported way to do this from user code.

## Tasks

- [x] Confirm whether `DoListItemsHaveFocus = true` / `CanListItemsLoseFocus = false` is the intended, supported way to start with an item focused.
- [x] Verify behavior end-to-end (a list that starts with item-level focus, navigable immediately by arrow keys/d-pad without an extra "enter" press).
- [x] If it works: document it in the Forms ListBox docs (keyboard/gamepad navigation section).
- [x] If it does not work or is awkward: fix it (ideally without a new property if the existing ones suffice), then document.

## Files

- `MonoGameGum/Forms/Controls/ListBox.cs`
- ListBox / Forms keyboard-navigation documentation

vchelaru:
Closes #3167.

## What this does

Research outcome: the feature the issue asks about **already works** — there is no API change needed. A `ListBox` can hand keyboard/gamepad focus directly to an item by setting `SelectedIndex` first and then `DoListItemsHaveFocus = true`, skipping the "focus the list, then press a key to enter it" step. This is confirmed by the existing `ListBox_DownArrowPressed_MovesSelectionDown` unit test. So this PR is **docs-only**.

## Changes

- **`docs/code/controls/listbox.md`** — new canonical **Keyboard and Gamepad Navigation** section. Explains the two-level focus model (top-level vs item-level), the `DoListItemsHaveFocus` / `CanListItemsLoseFocus` / `LoseListItemFocusOnPrimaryInput` properties, and the supported recipe for starting directly on an item — including the ordering gotcha (`SelectedIndex` must be set before `DoListItemsHaveFocus`).
- **`docs/code/events-and-interactivity/keyboard-support.md`** — adds a short ListBox navigation section (previously had none) that links back to the canonical section.
- **`docs/code/events-and-interactivity/gamepad-support.md`** — adds the link back, and **corrects an inaccuracy**: the page implied that focusing the ListBox alone lets the D-pad move the selection. In fact, at top-level focus the D-pad up/down tabs between controls (`HandleTab`); the user must press A to enter item-level navigation first. Prose now reflects this.

The focus model is modality-neutral, so it lives once on the ListBox page and the two input-support pages link to it rather than duplicating (avoids drift).

## ⚠️ XnaFiddle link needs a manual check

I added an XnaFiddle link to the new code sample on the ListBox page. The link was generated with the standard encoder and verified byte-identical to the encoder output, **but XnaFiddle generation has historically been flaky** — please click the link once and confirm it loads/compiles before merging. If it doesn't, removing that one line is a safe fix; the sample stands on its own without it.


vchelaru:
Follow-up fix (2nd commit): the first pass omitted `GumUI.UseKeyboardDefaults()`, so the ListBox samples couldn't actually respond to arrow keys — a ListBox reads the keyboard by iterating `FrameworkElement.KeyboardsForUiControl`, which stays empty until a keyboard is registered (unlike TextBox, which receives input just by being focused). Added the call to both samples plus a hint calling out that difference, and regenerated the XnaFiddle link to include it. The ⚠️ note above still applies — please click the (updated) fiddle once before merging.

PR #3228, opened 2026-06-20
Phase 0: coupling burn-down ratchet for tool UI/logic decoupling

### LINKED ISSUE #3225: Decouple Gum tool UI from logic — Phase 0+1 (instrument coupling + shallow WinForms scatter)
Tracking issue for the first execution chunk of the **Gum tool UI / logic decoupling** effort.

Source of truth (merged in #3224, do not relitigate):
- `Direction/ui-decoupling-plan.md` — the 6 phases + Definition of done
- `Direction/decisions/0003-decouple-tool-ui-from-logic.md` — decouple regardless of Avalonia; headless `net8.0` end-state so the compiler enforces the boundary
- `Direction/decisions/0004-viewmodels-expose-neutral-presentation-state.md` — ViewModels resolve display state in **neutral types** (`bool` visibility, Gum color type, `byte[]`/Gum image), framework types only in a thin converter/view layer

**Definition of done (every change):** code lands in a testable unit (service / ViewModel / utility) **and** a test — TDD when behavior changes, a characterization (pinning) test for behavior-preserving moves.

---

## Phase 0 — Instrument the coupling burn-down

Establish baseline metrics so progress is measurable and can't silently regress.

- [ ] Baseline count: `.Self` static references across the tool (`Gum/` etc.) — ~475 across ~103 files
- [ ] Baseline count: `System.Windows.*` usages in ViewModels — ~192 across ~18–35 VMs
- [ ] Baseline count: inline `new …Window()` / `MessageBox.Show` sites bypassing `IDialogService` — ~34 sites in ~16 files
- [ ] Baseline count: WinForms/WPF type leaks in "abstraction" interface signatures
- [ ] A **ratchet test** asserting each count ≤ current baseline (lower the number as items land). Not a hard "no `System.Windows` in VMs" ban — that fails immediately today.

## Phase 1 — Kill the shallow WinForms scatter & seal the leaky interface seams

Wide-but-shallow, low-risk. Each item gets a test per the Definition of done and is independently PR-able.

- [ ] **`Keys` enum → Gum-owned key enum** — `Gum/Managers/HotkeyManager.cs` (~48 uses), `Gum/Managers/DragDropManager.cs` (~43 uses)
- [ ] **Abstract drag-drop off WinForms types** (`DataObject`/`DragDropEffects`) — `DragDropManager.cs`
- [ ] **Move the one true WinForms `Form`** (`Gum/PropertyGridHelpers/Converters/FileListWindow.cs`) behind `IDialogService` as a WPF dialog
- [ ] **Delete vestigial state-tree WinForms glue** (`TreeNode`/`Keys` leftovers) — `StateTreeViewManager.cs`, `MainStatePlugin.cs`, `StateTreeViewRightClickService.cs`
- [ ] **Seal leaky "abstraction" interfaces** — `IGuiCommands.ShowSpinner()` returns a WPF `Spinner`; `ITabManager` takes `System.Windows.FrameworkElement` / `System.Windows.Forms.Control`
- [ ] **Inline dialogs → `IDialogService`** — ~34 sites (`DeleteLogic.cs`, `ProjectManager.cs`, `ElementTreeViewManager.RightClick.cs`, `ImportFromGumxView.xaml.cs`, …)

**Out of scope here (later phases):** the two big WinForms subsystems — `GraphicsDeviceControl` + input library (render host, Phase 4b) and `MultiSelectTreeView` + `ElementTreeViewManager` (Phase 4a); the `.Self` static drain (Phase 2); the headless-assembly extraction (Phase 3).

vchelaru:
## What & why

Phase 0 of the Gum tool **UI / logic decoupling** effort (see `Direction/ui-decoupling-plan.md`, ADR-0003, ADR-0004). It instruments the coupling burn-down so the effort is **measurable** and re-coupling **cannot be merged silently**.

This is a **source-scanning architecture guard** (a "ratchet"): a unit test that reads the tool's own `.cs` source, counts four kinds of UI-framework coupling, and asserts each count stays at-or-below a checked-in baseline. A hard "zero `System.Windows` in ViewModels" rule would fail on day one (~21 VMs import it today); the ratchet instead locks in "no worse than today" and is tightened as coupling is removed.

**Test-only — zero production tool code is touched.**

## What's in it (`Tool/Tests/GumToolUnitTests/Coupling/`)

- **`CouplingScanner.cs`** — the single canonical, deterministic counting method for the four metrics. No separate grep-based method that could diverge.
- **`CouplingScannerTests.cs`** — inline-sample unit tests that pin each counting rule (the executable spec).
- **`CouplingRatchetTests.cs`** — runs the scanner over the live `Gum/` tree, asserts each count ≤ baseline; a fixture **sanity floor** fails the guards loudly rather than passing vacuously if the scan ever points at a degenerate/empty root.
- **`coupling-baseline.json`** — the checked-in burn-down dashboard (the four numbers + the documented rule for each).

## The four baselines (today)

| Metric | Baseline |
|---|---|
| `.Self` statics (excluding sanctioned `ObjectFinder.Self`) | **194** |
| `System.Windows.*` usage in ViewModels | **146** |
| Inline dialog sites bypassing `IDialogService` | **7** |
| WinForms/WPF type leaks inside interface bodies | **11** |

## How Phase 1 interacts

Each Phase 1 PR **lowers the relevant baseline in the same diff that removes the coupling** — that one-line `coupling-baseline.json` edit is what makes the burn-down self-proving (you can't claim "decoupled X" without the number actually dropping).

## Notes

- **qa independently verified the core guarantee:** lowering a baseline and injecting synthetic coupling into a real `Gum/` file both turned the ratchet red with exact predicted deltas (then reverted). It cannot go red on the unmodified tree.
- **Two documented limitations are pinned by tests:** a metric-4 brace-matcher interpolation under-count (immaterial — Gum interfaces are signature-only) and that metric-2 `count == 0` is necessary-but-not-sufficient for "WPF-free" (a Phase 3 completion audit must also grep fully-qualified `System.Windows.` in VM files).
- The handoff's `~34` inline-dialog estimate was measured down to **7** (only 1 *live* `MessageBox.Show`) — Phase 1's dialog item is correspondingly smaller.

## Test plan

`GumToolUnitTests` Coupling suite: **19 passed / 0 failed**. Full suite previously **916 passed / 5 skipped / 0 failed**, no regressions.

```
dotnet test Tool/Tests/GumToolUnitTests/GumToolUnitTests.csproj --filter "FullyQualifiedName~GumToolUnitTests.Coupling"
```

Part of #3225. *(Intentionally does not auto-close — #3225 is the umbrella tracking issue for all of Phase 0 + Phase 1.)*

### DIFF
--- a/Tool/Tests/GumToolUnitTests/Coupling/CouplingRatchetTests.cs
+++ b/Tool/Tests/GumToolUnitTests/Coupling/CouplingRatchetTests.cs
@@ -1,0 +1,221 @@
+using System;
+using System.IO;
+using System.Text.Json;
+using Shouldly;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace GumToolUnitTests.Coupling;
+
+/// <summary>
+/// Locates the Gum tool source tree, runs <see cref="CouplingScanner"/> over it once, and
+/// loads the checked-in baseline. Shared across all ratchet tests via <see cref="IClassFixture{T}"/>
+/// so the (single) full-tree scan happens only once per test run.
+/// </summary>
+public sealed class CouplingScanFixture
+{
+    public CouplingMetrics Measured { get; }
+    public CouplingBaseline Baseline { get; }
+    public string ToolSourceRoot { get; }
+
+    public CouplingScanFixture()
+    {
+        string repoRoot = LocateRepoRoot();
+        ToolSourceRoot = Path.Combine(repoRoot, "Gum");
+
+        CouplingScanner scanner = new CouplingScanner();
+        Measured = scanner.ScanToolSource(ToolSourceRoot);
+
+        // Sanity floor: a found-but-degenerate tool root (empty, wrong directory, or a future
+        // source reorg) would make the scan return near-zero counts, and EVERY ratchet guard
+        // (measured <= baseline) would then pass VACUOUSLY -- 0 <= baseline is always true.
+        // Throwing here fails all ratchet tests loudly instead of going green for free. The
+        // floors sit well below today's real counts (~470 files / ~55 ViewModels) but far above
+        // zero, so they only trip on a genuinely broken scan, never on legitimate burn-down.
+        const int minimumFilesScanned = 100;
+        const int minimumViewModelFilesScanned = 10;
+        if (Measured.FilesScanned < minimumFilesScanned ||
+            Measured.ViewModelFilesScanned < minimumViewModelFilesScanned)
+        {
+            throw new InvalidOperationException(
+                $"Coupling scan looks degenerate: scanned {Measured.FilesScanned} files and " +
+                $"{Measured.ViewModelFilesScanned} ViewModel files under '{ToolSourceRoot}' " +
+                $"(floors: {minimumFilesScanned} files, {minimumViewModelFilesScanned} ViewModels). " +
+                "The scanner almost certainly pointed at the wrong root -- the ratchet guards would " +
+                "otherwise pass vacuously against an empty or incorrect tree.");
+        }
+
+        string baselinePath = Path.Combine(
+            repoRoot, "Tool", "Tests", "GumToolUnitTests", "Coupling", "coupling-baseline.json");
+        Baseline = LoadBaseline(baselinePath);
+    }
+
+    /// <summary>
+    /// Walks up from the test assembly's location until it finds the repo root, identified by the
+    /// presence of the Gum tool's main project file (<c>Gum/Gum.csproj</c>). This is robust for both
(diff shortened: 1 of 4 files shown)


vchelaru:
Closing — superseded by the decision to rely on the compiler boundary (the headless net8.0 assembly per ADR-0003) as the real enforcement, instead of a runtime coupling-counting ratchet. The ratchet largely duplicates what the compiler enforces for free once ViewModels/interfaces move into the no-WPF assembly; for the non-compiler-enforceable patterns (.Self, MessageBox.Show) a Roslyn banned-API analyzer is the lighter tool if a guard is ever needed. Baseline numbers are retained in notes.

PR #3247, opened 2026-06-21
#3233: Run Skia tests in CI, block on stable suites, probe raylib headless on macOS

### LINKED ISSUE #3233: CI does not run raylib or Skia test suites — regressions ship undetected
## Problem

`.github/workflows/build-and-test.yaml` explicitly **excludes the raylib and Skia test suites from CI** ("Bucket C — WILL NOT RUN IN CI", because GitHub runners have no GPU and these tests instantiate a real window / graphics device):

```
# Bucket C — WILL NOT RUN IN CI (require a real graphics device / window ...):
#   - RaylibGum.Tests                (raylib opens a window on init)
#   - SkiaGum.Tests                  (Skia pipeline wires through window plumbing ...)
```

So **no raylib and no Skia test ever runs on a PR.** The raylib/Skia runtimes have zero automated regression protection at merge time, even though `RaylibGum.Tests` alone has ~350 tests.

## Why this matters (concrete evidence)

- **PR #3232** fixed a raylib font regression (a cache-hit served an empty font, so all but the first text rendered nothing and `RelativeToChildren` stacks collapsed). The raylib font tests that *exist* would not have caught it (they disable caching, see below) — and even a correct test would never have run, because the suite is excluded from CI.
- **There are already-red raylib tests sitting on `main` right now**, undetected purely because of this gap (see the linked shape-stroke issue). That's the failure mode in action: a real defect, shipped, invisible.

## Secondary weakness

The suites that **do** run are marked `continue-on-error: true`, so even a detected failure does not fail the job / block the merge — failures are advisory only.

## Possible directions (not prescriptive)

- Split `RaylibGum.Tests` (and `SkiaGum.Tests`) via xUnit `[Trait]` into **headless-safe** (font measurement, layout, data, property plumbing — mostly CPU) vs **needs-window** (actual rendering / draw-call assertions). Run the headless subset in CI. Much of raylib's logic only needs a window because `MeasureTextEx` is gated behind `IsWindowReady()`.
- Investigate a headless GL path on the runner (Mesa `llvmpipe` software GL, or `xvfb` on a Linux job) so a hidden window can initialize.
- Or a self-hosted / nightly job on a runner with a display.
- Reconsider `continue-on-error: true` so a red test actually blocks.
- **Near-term floor** (cheap, do regardless): make "ran `RaylibGum.Tests` / `SkiaGum.Tests` locally" an explicit pre-merge checklist item for any PR touching those runtimes, since CI cannot.

Surfaced during the retrospective on PR #3232.

vchelaru:
Closes #3233.

## What this does

Closes the CI gap where raylib and Skia suites never ran on a PR, and makes the suites that *do* run actually block. Validated by this PR's own CI run (both `windows-latest` and `macos-15` green).

**Net effect on the workflow:** Skia is added as a blocking suite, the existing Bucket-A suites become blocking, and **no raylib step is added** — the raylib CI experiment below was run and reverted within this PR (see §3). raylib-on-CI is tracked separately in #3250.

### 1. `SkiaGum.Tests` now runs in CI as a **blocking** suite ✅
The old exclusion comment claimed Skia "wires through window plumbing that doesn't cooperate headless." That is **not true of these tests** — they render into an in-memory CPU raster `SKSurface`, with no GPU, window, or GL context. **Confirmed green on both Windows and macOS runners** this PR (189 tests). Moved to Bucket A.

### 2. Bucket-A suites now block the merge ✅
Removed `continue-on-error: true` from every Bucket-A suite (MonoGame, Sokol, GumToolUnitTests, Gum.Bundle, Gum.ProjectServices, Gum.Analyzers, Gum.Cli, + new Skia). A red test now **fails the job** instead of being advisory. `Publish Test Results` runs with `always()` so the report still publishes when a suite fails. **All of them passed on both runners this PR**, so the gate doesn't spuriously block.

- **Trade-off:** the first failing suite short-circuits the suites listed after it (the report still covers whatever ran). The alternative — keep `continue-on-error` + a final gate step on each suite's `outcome` — runs *all* suites every time at the cost of a second hand-maintained list. I went simple; happy to switch to the gate if you'd rather have full reporting on every run.

### 3. raylib in CI — experiment ran, decision is **don't (yet)**; tracked in #3250
Per the "experiment before splitting" decision, I temporarily added an advisory macOS step that ran `RaylibGum.Tests` to learn whether a hidden GLFW window can init on the runner. **Result: it hangs.** The native lib + GLFW backend loaded, then `InitWindow` blocked at window creation and hit the 10-minute timeout (swallowed by `continue-on-error`, so the job stayed green). GLFW/Cocoa window creation needs the **main thread + a window server**, and xUnit runs tests off the main thread — GL availability was never even reached.

So macOS is a dead end. **The probe was removed in this same PR — the final diff contains no raylib CI step.** raylib stays local-only (see §4); the finding is recorded in the workflow's Bucket C comment.

- **Next lever, tracked separately in #3250:** a **Windows runner + Mesa `llvmpipe`** — software GL 3.3 *and* Win32 window creation, which (unlike Cocoa) is not main-thread-coupled. The more promising path, but a separate, multi-iteration setup — out of scope here.

### 4. Commit-discipline floor for raylib ✅
Since raylib can't run in CI, running `RaylibGum.Tests` locally is the merge floor for raylib-touching work. Hardened the `gum-unit-tests` skill and added a rule to the `coder` agent (re-read every task).

## Why `--no-build` is safe
Both test projects are already in `AllLibraries.sln` with `Release|Any CPU.Build.0` set, so the existing "Build AllLibraries" step compiles them (incl. the net10 Skia project — proving the runners already have the .NET 10 runtime). The exclusion was purely the test-run allow-list, not a build gap.

## Heads-up
Flipping the Bucket-A suites to blocking means a future runner-only red (previously hidden by advisory mode) will block. This PR's run shows them all green today, so nothing is broken now — but it's the intended new behavior going forward.

## Deliberately not done here
- No headless/GL test split for raylib — the experiment showed the blocker is window creation, not test partitioning, so a split wouldn't help on macOS.
- No Windows + Mesa `llvmpipe` setup — tracked as the follow-up in #3250.


vchelaru:
Heads-up on the raylib-in-CI part of this PR: the advisory macOS probe described in §3 was **run and then reverted within this same PR** — the final diff adds **no raylib CI step**. On the macOS runner raylib's `InitWindow` hung at window creation (GLFW/Cocoa needs the main thread + a window server, xUnit runs off the main thread; GL availability was never even reached), so it's pure wasted runner time. raylib therefore stays **local-only** with the commit-discipline floor (§4). The remaining lever — a Windows runner + Mesa `llvmpipe` — is tracked separately in #3250. Updated the PR description to match.

PR #3252, opened 2026-06-21
#3250: Advisory CI probe — run RaylibGum.Tests on Windows via Mesa llvmpipe

### LINKED ISSUE #3250: raylib tests in CI: try Windows + Mesa llvmpipe (software GL) — macOS probe hung at window creation (#3233 follow-up)
## Background

Follow-up to #3233 / PR #3247. That work brought `SkiaGum.Tests` into CI (it renders into an in-memory CPU raster `SKSurface`, so it's fully headless) and made the stable suites block on failure — but **`RaylibGum.Tests` still does not run in CI.**

raylib's `InitWindow` needs an OpenGL 3.3 context the GitHub-hosted runners don't provide. As part of #3233 we ran an advisory probe of `RaylibGum.Tests` on the **macOS** runner to see whether a hidden GLFW window could init:

```
INFO: Initializing raylib 5.5
INFO: Platform backend: DESKTOP (GLFW)
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
INFO:     > raudio:.... loaded (optional)
```

…and then it **hung at window creation** and hit a 10-minute timeout. The native lib + GLFW backend load fine, but `InitWindow` blocks creating the window — consistent with GLFW/Cocoa requiring the **main thread + a window server**, while xUnit runs tests off the main thread. GL availability was never even reached. The probe was removed (it was pure wasted runner time); the finding is recorded in `build-and-test.yaml`'s Bucket C comment.

## Proposed next lever: Windows runner + Mesa `llvmpipe`

Windows looks more promising than macOS here:
- **Software GL 3.3** can be supplied by Mesa's `llvmpipe` (software rasterizer) — drop the `opengl32.dll` + `libgallium`/`libglapi` from a `mesa-dist-win` build next to the test host, with `GALLIUM_DRIVER=llvmpipe`.
- **Win32 window creation is not main-thread-coupled** the way Cocoa is, so the macOS hang shouldn't recur.

### Tasks
- [ ] Add a CI step (Windows job) that drops in Mesa `llvmpipe` and runs `RaylibGum.Tests` — start advisory (`continue-on-error`), separate results dir, with a `timeout-minutes` guard like the macOS probe had.
- [ ] Confirm `InitWindow` succeeds headless (the bootstrap already uses `ConfigFlags.HiddenWindow`) and the suite actually runs to completion.
- [ ] If green and stable, promote `RaylibGum.Tests` to a blocking Bucket-A step and update the allow-list comment in `build-and-test.yaml`, the `gum-unit-tests` skill, and the `coder` agent (drop the "raylib does not run in CI" floor).

### Alternatives if Windows + Mesa doesn't pan out
- `xvfb` + Mesa on an added **Linux** job (the conventional headless-GL CI recipe).
- A self-hosted / nightly runner with a real display.
- Keep raylib local-only and lean on the commit-discipline floor (current state) — close as "not worth the runner cost."

## Acceptance
`RaylibGum.Tests` runs in CI on PRs (ideally blocking), OR this is closed with the decision recorded that local-only is the accepted floor.

Refs: #3233, PR #3247.

vchelaru:
## What

Makes `RaylibGum.Tests` run in CI as a **blocking** suite on the Windows job, using Mesa's `llvmpipe` software OpenGL to supply the GL 3.3 context the GPU-less runners lack.

Shipped in two commits so the risky part was proven before it could gate anything:
1. **Advisory probe** — `continue-on-error`, separate results dir, timeout guard. Result: **351/351 raylib tests passed headless**, ~12 s incremental on the Windows job (7 s Mesa download + 5 s test run).
2. **Promotion to blocking** — dropped `continue-on-error`, routed the trx into the gating `TestResults` reporter, hardened the Mesa download with a retry (it now feeds a blocking gate), and lifted the raylib local-run merge floor.

## Why

raylib's `InitWindow` needs an OpenGL 3.3 context the GitHub runners don't provide, so `RaylibGum.Tests` was CI-excluded (the #3233 macOS probe *hung* at GLFW/Cocoa window creation). On Windows:
- **Software GL 3.3** comes from Mesa's `llvmpipe` rasterizer.
- raylib's native lib imports `opengl32.dll` *by name*, and Windows resolves that from the test host's own directory **before** System32 — so dropping Mesa's `opengl32.dll` (+ gallium/glapi siblings) next to the test binaries overrides the GL-less system stub.
- **Win32 window creation is not main-thread-coupled** like Cocoa, so the macOS hang doesn't recur.

Mesa is a **pinned** `pal1000/mesa-dist-win` release (`26.1.2`).

## Why Windows and not a Linux/Xvfb job

The conventional headless-GL recipe is Linux + Xvfb + Mesa, and it was considered. But the incremental cost on Windows turned out to be only **~12 s** (the raylib build already happens in "Build AllLibraries"; only the Mesa download + 1 s test run are new). A separate Linux job would save ~12 s of wall-clock at the cost of a whole new job to build and maintain (apt Mesa/Xvfb, duplicate compile) — not worth it. The Windows shim empirically works clean, so it wins on simplicity.

## Floor docs updated (the local-run requirement is retired)

A green CI run now covers raylib, so the "you must run `RaylibGum.Tests` locally before merging raylib changes" floor is dropped from the `gum-unit-tests` skill, the `coder` agent, and the Bucket C → Bucket A move in the allow-list comment.

## Verification

This is CI config; the verification is the CI run itself. The advisory commit already showed 351/351 green; the blocking commit re-runs the same suite as a gating step.

Closes #3250


vchelaru:
Advisory probe (commit 1) ran clean on the windows-latest job: **`Passed! - Failed: 0, Passed: 351, Skipped: 0, Total: 351, Duration: 1 s - RaylibGum.Tests.dll`**. Mesa install = 7 s (download `mesa3d-26.1.2` + copy `opengl32.dll`/`libgallium_wgl.dll`), test run = 5 s — ~12 s incremental on the Windows job. Commit 2 promotes it to a blocking Bucket-A step (no `continue-on-error`, trx in the gating reporter) and lifts the raylib local-run floor. Watching the blocking run now.

PR #3361, opened 2026-06-25
Document DeleteOptionsWindow's headless/ADR-0005 coupling in gum-tool-dialogs skill

vchelaru:
Adds a "Headless / ADR-0005 implication" note to the `gum-tool-dialogs` skill.

The skill already documented the two dialog systems and *why* `DeleteOptionsWindow` is standalone (plugins inject controls into a live WPF `StackPanel`, which the MVVM template system can't express). What it was missing — and what surfaced during the ADR-0005 headless-presentation refactor — is the decoupling consequence:

- The MVVM `DialogService` / `IDialogService` path is already headless, so dialogs on it don't couple callers to WPF.
- `DeleteOptionsWindow` does, and it is the specific coupling that blocks relocating `DeleteLogic` into the headless `Gum.Presentation` assembly.
- The coupling is the extension point itself (`PluginManager.ShowDeleteDialog`/`DeleteConfirmed` hand plugins a concrete WPF `StackPanel`). Undoing it is a plugin-facing API change — replace panel-injection with a data-driven options model — not a mechanical move.

Docs-only change to a checked-in skill file; no code touched.

### DIFF
--- a/.claude/skills/gum-tool-dialogs/SKILL.md
+++ b/.claude/skills/gum-tool-dialogs/SKILL.md
@@ -35,6 +35,8 @@ Used by: delete confirmation only (`DeleteLogic.ShowDeleteDialog`).
 
 **Not managed by DialogService** — no view model, no template selection, no attached property binding. Changes to `DialogWindow.xaml` or `Dialog.cs` have **zero effect** on this window.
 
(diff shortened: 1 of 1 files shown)


vchelaru:
Closing — this baked in-flight ADR-0005 migration state (what is headless now, what currently blocks DeleteLogic's relocation, what is left) into a persistent skill file, which rots the moment the migration moves. The durable fact (why DeleteOptionsWindow is a standalone code-behind window: runtime plugin StackPanel injection that the MVVM template system can't express) was already in the skill. The forward-looking decoupling direction belongs in the ADR, not the skill.

Issue #3293, opened 2026-06-23
Animation keyframe errors should be detected per-element (headless) so the tree '!' shows on project open and on edits, not just on selection


vchelaru:
## Problem

Animation "keyframe references a missing state/animation" errors are detected **only** through the State Animation plugin's per-selection view model. The plugin loads a `.ganx` only for the **currently selected** element, so these errors are invisible everywhere else:

- They do **not** appear on **project open** — `RefreshErrorIndicatorsForAllElements` runs, but the animation plugin's view model is empty, so no animation errors are reported for any element. Core/structural errors show at open; animation errors don't.
- They do **not** refresh **when you cause one** — e.g. remove or rename a state a keyframe references. The edit path mangles the view model before detection (see #3383), so the error only appears after re-selecting the element (which rebuilds the view model from disk).

> Originally filed as the symptom: *"tree-view `!` is stale — an animation frame referencing a missing state shows in the Errors tab but not on the element node until an unrelated tree refresh."* Investigation traced it to the above root cause; reframed here. A selection-refresh workaround was prototyped and rejected as a bandaid — it isn't needed once detection is selection-independent.

## Desired behavior

1. **On project open**, every element whose animation references a missing state/animation shows the red `!` on its tree node, at a glance.
2. **When you cause such an error** (remove/rename a referenced state), the `!` updates **immediately** on the node and in the Errors tab — no re-select needed.

## Approach

Move the keyframe → missing-state (and → missing-animation) validation into the **headless per-element checker** (`HeadlessErrorChecker.GetErrorsForInternal`), alongside the existing core checks. Feed animation data through an **injected provider** so the checker never touches the filesystem directly:

- Pure check: validate an `ElementAnimationsSave`'s keyframes against the `ElementSave`'s states/categories (and referenced animations) → `ErrorResult`s with a `GUM00xx` code.
- Tool provider: load each `.ganx` once into an in-memory cache, invalidate on save / file-change. State edits (the usual way you cause an error) don't change the `.ganx`, so they don't even invalidate — the check compares the cached keyframes against the live (in-memory) element states.
- CLI provider: read disk once per `gumcli check` run.

Consequences:
- (1) runs for all elements on project open; (2) runs per element on existing edit refreshes.
- Reads from disk/cache, so the indicator is correct even while the view model is mid-edit (robust to #3383's view-model mangling).
- `gumcli check` / CI catches dangling animation references for free.
[...]

PR #3393, opened 2026-06-27
Rename a state: update animation keyframe references and clear stale error (#3383)

### LINKED ISSUE #3383: Renaming a state doesn't update animation keyframe references (leaves a stale missing-state error)
## Summary

Renaming a state does **not** reliably update animations that reference that state. The animation keyframe keeps pointing at the old state name, producing a `Keyframe at time … references a state X which does not exist` error that lingers even though the rename should have cascaded to the animation.

## Steps to reproduce

1. Open a Screen or Component; add a state (optionally inside a category).
2. Open the Animations window, create an animation with a keyframe referencing that state.
3. Rename the state.
4. Observe: the animation now reports a missing-state error — the keyframe still references the old name.

## Expected

Renaming a state updates every animation keyframe that references it (both in memory **and** persisted to the `.ganx`), so no dangling missing-state error is produced.

## Technical findings

- `RenameLogic.RenameState` (`Gum/Logic/RenameLogic.cs`) updates *variable* references via `GetChangesForRenamedState` → `ApplyStateReferences`, but animation keyframe references are handled separately, by the animation plugin.
- **Ordering bug** in `MainStateAnimationPlugin.HandleStateRename` (`Gum/StateAnimationPlugin/MainStateAnimationPlugin.cs`):
  ```csharp
  RefreshViewModel();   // recomputes errors against the OLD keyframe names AND broadcasts RequestErrorRefreshMessage
  _renameManager.HandleRename(stateSave, oldName, _viewModel);   // only now rewrites keyframe.StateName
  ```
  Errors are computed/broadcast **before** the keyframe names are rewritten, and are never recomputed afterward — so a stale missing-state error remains.
- `RenameManager.HandleRename(StateSave, oldName, viewModel)` (`Gum/StateAnimationPlugin/Managers/RenameManager.cs`) rewrites `keyframe.StateName` in the in-memory viewmodel but — unlike the `ElementSave` overload — does **not** save the `.ganx`, so even the corrected reference may not persist across a reload.
- Possible secondary bug in that same method: `prefix = parentCategory + "/"` relies on `StateSaveCategory.ToString()` returning the category name, while the rest of the code keys off `category.Name`. If those differ, categorized-state renames won't match keyframes at all.

## Suggested fix direction

Rewrite keyframe references **before** errors are recomputed/broadcast, recompute errors after the rewrite, and persist the `.ganx`. Audit the category-prefix construction (`parentCategory + "/"` vs `category.Name + "/"`).

## Related

Surfaced while fixing #3293 (stale tree-view `!` error indicator). The indicator fix makes this lingering error visible on the tree node too, so the two are related but distinct.

vchelaru:
## Summary

Renaming a state did not reliably update animations that referenced it: the keyframe kept pointing at the old name, leaving a lingering `Keyframe at time … references a state X which does not exist` error (surfaced on the tree node and Errors tab after #3293/#3382). Fixes #3383.

The pre-fix symptom was deceptive: the error *vanished* the instant you renamed (the WPF binding nulled `StateName`, which trips neither error branch), so it looked fixed — but the keyframe was never actually rewritten, and re-selecting/reloading the element rebuilt the original reference and the error returned.

## Root causes addressed

1. **Plugin ordering** — `MainStateAnimationPlugin.HandleStateRename` called `RefreshViewModel()` (which recomputes/broadcasts errors against the still-old keyframe names) **before** `RenameManager` rewrote the keyframes, and never recomputed afterward. Reordered: rewrite first, then `RefreshViewModel` recomputes + rebroadcasts.
2. **WPF binding nulled the reference** — rebuilding `AvailableStates` removed the old categorized name, so the keyframe ComboBox's two-way `SelectedItem` → `StateName` binding cleared `StateName` to null before the rewrite could match it. The post-rename name is now seeded into `AvailableStates` *before* the rewrite, so the binding accepts the rewritten value instead of nulling it.
3. **Prefix via `ToString()`** — the category prefix was built from `StateSaveCategory.ToString()` instead of `.Name`; switched to `.Name` to match the rest of the keyframe-name handling.

## Persistence — single save path (no double write)

`RenameManager.HandleRename(StateSave)` rewrites the keyframe and does **not** save. Assigning `keyframe.StateName` raises `PropertyChanged`, which the plugin already persists through its centralized `HandleDataChange` (`AnyChange`) save — the same single path the sibling `InstanceSave` overload relies on. An explicit save here would write the `.ganx` twice per rename and trip external (and Gum's own) file watchers on a change that already settled. The `ElementSave` overload still saves explicitly because renaming the element *file* changes no keyframe property, so nothing triggers the implicit save there.

## Tests

- `AnimationStateRenameErrorTests`: promoted `HandleRename_OfCategorizedState_RewritesKeyframeAndClearsErrorOnRefresh` to the #3383 ordering regression guard (rewrite-then-refresh leaves no stale error) and removed the now-fixed `RenameSequence_AsThePluginRunsIt_LeavesStaleError` characterization.
- `RenameManagerTests`: the existing `HandleRename_State_RenamesMatchingKeyframeStateName` / `…StateCategory…` already pin the rewrite.

All 18 `StateAnimationPlugin` tests green; `GumFull.sln` builds with 0 errors.

## Manual verification

The WPF ComboBox-binding half can't be unit-tested (needs a WPF host), so it's verified in the running tool: add a state (in a category), create an animation keyframe referencing it, rename the state → no missing-state error appears, the keyframe shows the new name, and it persists across reselect/reload.

Closes #3383

### DIFF
--- a/Gum/StateAnimationPlugin/MainStateAnimationPlugin.cs
+++ b/Gum/StateAnimationPlugin/MainStateAnimationPlugin.cs
@@ -267,12 +267,29 @@ public class MainStateAnimationPlugin : PluginBase
 
     private void HandleStateRename(StateSave stateSave, string oldName)
     {
-        RefreshViewModel();
-
-        if (_selectedState.SelectedElement != null && _viewModel != null)
+        var element = _selectedState.SelectedElement;
+        if (element != null && _viewModel != null)
         {
+            // RenameLogic renames the state in the model BEFORE notifying plugins, so stateSave
+            // already holds the new name. Rewrite the keyframes referencing the old name BEFORE
+            // RefreshViewModel rebuilds AvailableStates. Seed the post-rename name into the existing
+            // AvailableStates first so the keyframe ComboBox's two-way SelectedItem binding accepts
+            // the rewritten StateName instead of nulling the (now-absent) old selection — that null
+            // is what dropped the reference and left a stale missing-state error. (#3383)
+            var category = element.Categories.FirstOrDefault(item => item.States.Contains(stateSave));
+            var newName = category != null ? category.Name + "/" + stateSave.Name : stateSave.Name;
+            if (!AvailableStates.Contains(newName))
+            {
+                AvailableStates.Add(newName);
+            }
+
             _renameManager.HandleRename(stateSave, oldName, _viewModel);
         }
+
+        // Rebuild AvailableStates from the model and recompute errors against the rewritten
+        // keyframes; CreateViewModel broadcasts RequestErrorRefreshMessage so the Errors tab and
+        // the tree "!" indicator re-check.
+        RefreshViewModel();
     }
 
     private void HandleStateAdd(StateSave state)
--- a/Gum/StateAnimationPlugin/Managers/RenameManager.cs
+++ b/Gum/StateAnimationPlugin/Managers/RenameManager.cs
@@ -127,11 +127,14 @@ namespace StateAnimationPlugin.Managers
             string prefix = "";
             if(parentCategory != null)
             {
-                prefix = parentCategory + "/";
+                // Use .Name explicitly rather than relying on StateSaveCategory.ToString() returning
+                // the name — the rest of the keyframe name handling keys off category.Name.
+                prefix = parentCategory.Name + "/";
             }
 
             oldName = prefix + oldName;
 
+            bool didRename = false;
             foreach(var animation in viewModel.Animations)
             {
                 foreach(var keyframe in animation.Keyframes)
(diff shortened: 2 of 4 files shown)


vchelaru:
Closing as redundant. #3383 was already fixed at the root by #3389 (merged onto the same base this branch was cut from, mid-session): the keyframe state ComboBox was made IsEditable with a Text binding, eliminating the StateName null-coercion entirely, plus the HandleStateRename ordering fix. This PR re-introduced the AvailableStates-seeding workaround that #3389 deliberately reverted, so merging it would regress that design.

Issue #3434, opened 2026-07-02
RaylibGum: render-to-texture (render target) support for containers


vchelaru:
Part of #3432 (RaylibGum feature-parity umbrella).

## Problem

RaylibGum has no general render-to-texture (render target) support for containers. `ContainerRuntime.IsRenderTarget` compiles on raylib but is never read by the raylib renderer — `Renderer.Draw` / `DrawGumRecursively` (`Runtimes/RaylibGum/RenderingLibrary/Renderer.cs:126-289`) walk the tree drawing directly to the backbuffer, with clipping done via scissor (`BeginScissorMode`) only. There is no offscreen-bake pass. Contrast the MonoGame path (`RenderingLibrary/Graphics/Renderer.cs`), which bakes `IsRenderTarget` containers into offscreen targets before compositing.

The primitives already exist on raylib — this is wiring, not new capability. `ShadowBlurRenderer.cs` already renders to `RenderTexture2D` and manages those targets through `RenderTextureService`, a subclass of the **same shared `RenderTargetServiceBase<T>`** MonoGame's `RenderTargetService` uses (`RenderingLibrary/Graphics/RenderTargetService.cs`). Its own class doc names the intended future consumers: *"future raylib offscreen-rendering consumers (RT-based clipping, post-effects) should each own their own instance keyed by the renderable that owns the offscreen pass."*

**Scope:** this issue is render targets ONLY. The `#816`-style per-container post-process shader (`ContainerRuntime.RenderTargetEffect` / `SourceShaderFile`, currently `#if XNALIKE`) is explicitly **out of scope** and tracked as a separate parity item — though it builds on top of this one.

## Work items

- [ ] **1. Render-to-texture on containers.** When a container has `IsRenderTarget == true`, bake its subtree into an offscreen `RenderTexture2D` (`BeginTextureMode` → draw children → `EndTextureMode`), cache it keyed by the renderable, then draw that texture to the screen in place of walking the children live. Add the bake pass to `DrawGumRecursively` (`Renderer.cs:238-289`); today the tree is drawn directly with no such pass.

- [ ] **2. Blend / alpha on the composited container.** Properties that on MonoGameGum were only meaningful *through* a render target — group-level alpha/opacity and blend applied to the whole container as a unit — should be supported and unit-tested on raylib. Applied when the cached RT is drawn back to the screen (`DrawTexture*` with the container's blend/tint), mirroring the MonoGame composite. Verify these actually change output, not just compile.

[...]

PR #3473, opened 2026-07-03
raylib: implement full Blend mapping (Replace/ReplaceAlpha/SubtractAlpha/MinAlpha)

### LINKED ISSUE #3470: raylib: Blend modes collapse to Alpha (Replace/ReplaceAlpha/SubtractAlpha/MinAlpha unmapped)
## Problem

`Runtimes/RaylibGum/Renderables/BlendModeExtensions.cs:13-17` only maps `Blend.Additive`; every other Gum `Blend` value (`Replace`, `ReplaceAlpha`, `SubtractAlpha`, `MinAlpha`) falls through to `BlendMode.Alpha` — a silent, incorrect visual result. Affects `Sprite` (`Sprite.cs:167`) and `NineSlice` (`NineSlice.cs:227`), the two call sites that read `Blend.Value.ToRaylibBlendMode()`.

Split off from the umbrella parity tracking issue #3432 (P1 item "Blend modes collapse to Alpha").

## Why this is not blocked

raylib-cs's `BlendMode` enum has no canned entries for Gum's alpha-only/replace modes, but raylib exposes the low-level `Rlgl.SetBlendFactorsSeparate(srcRGB, dstRGB, srcAlpha, dstAlpha, eqRGB, eqAlpha)` + `Raylib.BeginBlendMode(BlendMode.CustomSeparate)` API — arbitrary separate color/alpha blend factors and equations. This repo already uses exactly that mechanism for two ad-hoc blend passes in `Runtimes/RaylibGum/RenderingLibrary/BatchDrawCallCounter.cs:130-185` (the render-target premultiply pass and the additive-premultiplied composite).

Better still, Gum already has a backend-agnostic `BlendState` model (`ToolsUtilities/Common/BlendState.cs`) with `Blend`/`BlendFunction` enums (`One`, `Zero`, `SourceAlpha`, `InverseSourceAlpha`, ...; `Add`, `ReverseSubtract`, `Min`, ...) that map 1:1 onto standard GL blend constants. `Gum.RenderingLibrary.BlendExtensions.ToBlendState(...)` (`Gum/RenderingLibrary/BlendExtensions.cs:16-38`) already turns each Gum `Blend` value into one of these — including `BlendState.SubtractAlpha`/`ReplaceAlpha`/`MinAlpha`/`Opaque` — for MonoGame. That method already compiles under `RAYLIB` (guarded by `#if !NO_XNA && !SKIA`, and RaylibGum defines neither).

## Suggested improvement

Reuse `blend.ToBlendState(isUsingPremultipliedAlpha: false)` (raylib currently has no premultiplied-alpha texture handling, so `false` matches existing behavior) and translate its abstract `Blend`/`BlendFunction` fields to GL int constants via a small lookup table (most constants — `GlOne`, `GlSrcAlpha`, `GlOneMinusSrcAlpha`, `GlFuncAdd` — already exist in `BatchDrawCallCounter.cs`; needs `GlZero`, `GlFuncReverseSubtract`, `GlFuncMin` added), then call `SetBlendFactorsSeparate` + `BeginBlendMode(CustomSeparate)`. This reuses the exact semantics MonoGame already has, giving parity rather than a guessed GL approximation. `Normal`/`Additive` can keep the existing simple `BlendMode.Alpha`/`BlendMode.Additive` path since those already match.

## Source

- `Runtimes/RaylibGum/Renderables/BlendModeExtensions.cs:7-18` — the mapping to fix.
- `Runtimes/RaylibGum/Renderables/Sprite.cs:189`, `Runtimes/RaylibGum/Renderables/NineSlice.cs:227` — call sites.
- `Runtimes/RaylibGum/RenderingLibrary/BatchDrawCallCounter.cs:130-185` — existing precedent for the `SetBlendFactorsSeparate`/`CustomSeparate` mechanism.
- `Gum/RenderingLibrary/BlendExtensions.cs:16-38`, `ToolsUtilities/Common/BlendState.cs` — the reusable backend-agnostic blend model.

vchelaru:
## Summary

- `Runtimes/RaylibGum/Renderables/BlendModeExtensions.cs` only mapped `Blend.Additive`; every other Gum `Blend` value (`Replace`, `ReplaceAlpha`, `SubtractAlpha`, `MinAlpha`) silently fell through to `BlendMode.Alpha`.
- Reuses the existing backend-agnostic `Gum.BlendState` model (the same one MonoGame's `BlendExtensions.ToBlendState` turns into XNA `BlendState` objects) and maps its abstract blend-factor/equation fields onto raylib's low-level `Rlgl.SetBlendFactorsSeparate` + `BlendMode.CustomSeparate` — the same mechanism this codebase already uses for the render-target premultiply/additive-premultiplied passes (`BatchDrawCallCounter.cs`). This gives raylib parity with MonoGame's blend semantics rather than an approximation.
- `Sprite`, `NineSlice`, and (as a side effect of the shared `BeginBlendMode(Blend)` overload) `RectangleRuntime`'s raylib `LineRectangle` backing all pick up the fix for free — one shared mapping, three call sites.
- Updated stale doc comments describing the old "only Additive maps" limitation (`LineRectangle.cs`, a `SkiaGum.Tests` cross-reference comment).

## Test plan
- [x] New pixel-readback tests (`Tests/RaylibGum.Tests/Rendering/BlendModeRenderTests.cs`) prove each of the 4 previously-broken modes produces its distinct, correct GPU result — verified red against the pre-fix mapping (via `git stash`) and green against the fix.
- [x] Extended the existing draw-call-count regression test (`RendererDrawCallMatrixTests.BlendedSprite_RendersAndIsCountedAcrossBlendFlushes`) to a `[Theory]` covering all 6 `Blend` values, confirming every mode still routes through the counted `BeginBlendMode` wrapper.
- [x] Full `RaylibGum.Tests` suite green (409/409), no regressions.
- [x] Added the alpha-only blend row to `Samples/raylib/Screens/SpriteScreen.cs`, mirroring the MonoGame sample (which already had it) — removed the stale "raylib can't do this" comment.
- [ ] Manual: Visual Studio is open on `Samples/raylib/RaylibSample.sln`. Run it, click the **Sprite** nav button, and check the two blend rows near the bottom:
  - "Blend, normal rendering (Normal, Additive, Replace)" — the Replace cell should now look like a flat, un-blended bear (no background bleed-through), visibly different from Normal.
  - "Blend, alpha-only on render-target Container (SubtractAlpha, ReplaceAlpha, MinAlpha)" — a new row; SubtractAlpha punches a transparent bear-shaped hole in the red square, ReplaceAlpha/MinAlpha show the bear silhouette at differing opacity.

Closes #3470

### DIFF
--- a/MonoGameGum.Tests/RenderingLibraries/BlendExtensionsTests.cs
+++ b/MonoGameGum.Tests/RenderingLibraries/BlendExtensionsTests.cs
@@ -47,4 +47,33 @@ public class BlendExtensionsTests : BaseTestClass
     {
         Gum.BlendState.SubtractAlpha.ToBlend().ShouldBe(Blend.SubtractAlpha);
     }
+
+    // Forward direction (Blend -> BlendState). Previously untested even though ToBlendState is what
+    // every runtime (SpriteRuntime, ContainerRuntime, NineSliceRuntime, TextRuntime, ...) actually
+    // calls to apply a Blend value — RaylibGum's BlendModeExtensions (issue #3470) now also depends
+    // on it to derive raylib's low-level blend factors, so pin the exact BlendState each Blend value
+    // resolves to.
+    [Fact]
+    public void ToBlendState_ReturnsMinAlpha_WhenBlendIsMinAlpha()
+    {
+        Blend.MinAlpha.ToBlendState().ShouldBeSameAs(Gum.BlendState.MinAlpha);
+    }
+
+    [Fact]
+    public void ToBlendState_ReturnsOpaque_WhenBlendIsReplace()
+    {
+        Blend.Replace.ToBlendState().ShouldBeSameAs(Gum.BlendState.Opaque);
+    }
+
+    [Fact]
+    public void ToBlendState_ReturnsReplaceAlpha_WhenBlendIsReplaceAlpha()
+    {
+        Blend.ReplaceAlpha.ToBlendState().ShouldBeSameAs(Gum.BlendState.ReplaceAlpha);
+    }
+
+    [Fact]
+    public void ToBlendState_ReturnsSubtractAlpha_WhenBlendIsSubtractAlpha()
+    {
+        Blend.SubtractAlpha.ToBlendState().ShouldBeSameAs(Gum.BlendState.SubtractAlpha);
+    }
 }
--- a/Runtimes/RaylibGum/Renderables/BlendModeExtensions.cs
+++ b/Runtimes/RaylibGum/Renderables/BlendModeExtensions.cs
@@ -2,18 +2,114 @@ using Raylib_cs;
 
 namespace Gum.Renderables;
 
+/// <summary>
+/// Translates Gum's backend-agnostic <see cref="global::Gum.RenderingLibrary.Blend"/> values into
+/// what raylib needs to reproduce them. <see cref="global::Gum.RenderingLibrary.Blend.Normal"/> and
+/// <see cref="global::Gum.RenderingLibrary.Blend.Additive"/> have a direct raylib
+/// <see cref="BlendMode"/> equivalent; every other value (<c>Replace</c>, <c>ReplaceAlpha</c>,
+/// <c>SubtractAlpha</c>, <c>MinAlpha</c> — issue #3470) needs distinct per-channel blend
+/// factors/equations that only <c>Rlgl.SetBlendFactorsSeparate</c> +
+/// <see cref="BlendMode.CustomSeparate"/> can express.
+///
+/// <para>Rather than re-deriving those factors, this reuses
+/// <see cref="global::Gum.RenderingLibrary.BlendExtensions.ToBlendState"/> — the same
+/// backend-agnostic <see cref="global::Gum.BlendState"/> that MonoGame turns into XNA
+/// <c>BlendState</c> objects — and maps its abstract <see cref="global::Gum.Blend"/> /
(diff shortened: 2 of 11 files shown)


vchelaru:
**Follow-up fix (2026-07-03):** manual visual comparison against the MonoGame sample (side-by-side, both `SpriteScreen`s open) caught a real bug the pixel tests initially missed: they only read the render-target cell's *raw bake*, not what raylib actually displays after compositing it. SubtractAlpha showed magenta instead of revealing the background behind the punched hole; ReplaceAlpha/MinAlpha/Replace had the same class of issue less visibly.

Root cause: raylib's render-target bake stores an *already-premultiplied* result and composites it with `BlendMode.AlphaPremultiply` (issue #3434's dark-fringe fix). The color factors reused directly from MonoGame's `BlendState` (which assumes non-premultiplied storage, since MonoGame's own RT composite re-multiplies by alpha) leave color untouched while alpha changes — valid for MonoGame, invalid for raylib's premultiplied pipeline, so leftover full-intensity color bled through at composite time.

Fixed with a raylib-specific, premultiplied-consistent color-factor derivation used only while an active bake is in progress (falls back to the original MonoGame-matching factors for a plain Sprite drawn straight to the screen, where nothing further composites the result). Rewrote the pixel tests to assert on the *composited* result (nested outer render target), not just the raw bake — confirmed red against the old derivation, green against the new one.

PR #3540, opened 2026-07-08
Fix/android monogame problem

KMemphis:


vchelaru:
Not sure what this PR is all about, but there are 30 files changed, that seems wrong - should there be this many changes?

vchelaru:
Thanks for this! The Android element-load fix here already merged as #2471 back in April 2026 — `ReadAndDetectFormat` on `main` has used `FileManager.FromFileText` since then, so the runtime file-loading bug is fixed. The only new change in this PR is a stray File-menu item ("Yotsuba Hybrid") that looks like leftover debug code, and the branch is based on a pre-#2471 `main` so it conflicts. Closing as a duplicate — appreciate the contribution and the original find.

Issue #3615, opened 2026-07-11
Unify CustomSetPropertyOnRenderable across MonoGame/Raylib/Skia (per-run BBCode font-tag resolution has diverged into 3 architectures)


vchelaru:
## Problem

`CustomSetPropertyOnRenderable.cs` is one of the last unconverted cross-platform files — only MonoGame/KNI/FNA file-link the shared `Gum/Wireframe/CustomSetPropertyOnRenderable.cs`; Raylib (`Runtimes/RaylibGum/Renderables/CustomSetPropertyOnRenderable.cs`, ~900 lines) and Skia (`Runtimes/SkiaGum/CustomSetPropertyOnRenderable.cs`, ~1290 lines) each carry their own divergent copy against the ~2430-line shared source. This is already tracked as a known convergence target (`.claude/designs/runtime-unification/RuntimeUnificationAndRefactor.md:166-176`, `RuntimeNorthStar.md:35`), flagged as needing "a dedicated research/design phase before convergence."

Investigating issue #3528 (add a BBCode `[HasShadow]` tag for inline drop-shadow) surfaced that the divergence is worse than a simple line-diff: the three backends have grown **different architectures** for the same feature — per-run baked-font BBCode tags (`[FontSize]`, `[OutlineThickness]`, `[IsBold]`, `[IsItalic]`):

- **MonoGame/KNI/FNA** resolves these through stack-based push/pop state (`fontNameStack`, `fontSizeStack`, `outlineThicknessStack`, etc.) inside `ApplyFontVariables`/`GetAndCreateFontIfNecessary`, living *inside* `Gum/Wireframe/CustomSetPropertyOnRenderable.cs` (~line 1122+).
- **Raylib** has none of that stack machinery in its `CustomSetPropertyOnRenderable.cs` copy. It independently reimplemented per-run `[FontSize]` swapping a different way, computing an ad hoc swap font directly inside `Runtimes/RaylibGum/Renderables/Text.cs` (`hasSwapFont` logic, ~lines 941-989).
- **Skia** has neither mechanism — it does not support per-run `[FontSize]` swapping at all yet.

Net effect: any new per-run baked-font BBCode tag (like `[HasShadow]`) currently requires three bespoke implementations in three different styles, rather than one shared implementation plus `#if` branches. Every feature in this class pays this tax until the file converges.

## Scope

[...]

PR #3628, opened 2026-07-12
Converge Sprite-family dispatch in CustomSetPropertyOnRenderable (Raylib <-> MonoGame/KNI/FNA)

vchelaru:
## Summary

Part of #3615 (Raylib/MonoGame runtime-file unification). This is the "Sprite family" convergence — disjoint from the font-region convergence in #3627 (different methods, no overlap).

Converges the two remaining per-platform copies of `CustomSetPropertyOnRenderable`'s Sprite-handling methods:
- `Gum/Wireframe/CustomSetPropertyOnRenderable.cs` (home copy — compiled into MonoGame/KNI/FNA and FRB)
- `Runtimes/RaylibGum/Renderables/CustomSetPropertyOnRenderable.cs` (Raylib copy)

Target methods: `TrySetPropertyOnSprite`, `AssignSourceFileOnSprite`, `GetAnimationChainList` (already identical, untouched). After this change the cross-file diff for these three methods is zero except for lines under mirrored `#if RAYLIB` / `#if !RAYLIB`.

### Structural changes (behavior-preserving)
- Home's `if/else if` chain converged onto Raylib's `switch` form (cleaner, per prior convergence precedent).
- `AssignSourceFileOnSprite`'s texture-load error handling unified onto the fuller structure (nullable `texture` + `loadException`, single `Texture2D` type reference instead of a redundant fully-qualified `Microsoft.Xna.Framework.Graphics.Texture2D`).
- Removed a dead `SpriteRuntime? asSpriteRuntime` local in the Raylib copy (unused; the disabled cases it was for are now live and each resolve their own runtime reference).
- Removed Raylib's `case nameof(Sprite.Texture): sprite.Texture = (Texture2D)value;` — redundant with the reflection fallback in `SetPropertyOnRenderable`, which already assigns the same value to the same renderable property when no case handles it. Home never had this case.

### Platform-necessary divergences (kept under mirrored `#if`)
- `Sprite.Color`: XNA/System.Drawing color handling is `#if !RAYLIB`-only. Raylib's branch stays absent — assigning `"Color"` via the state/variable system remains unimplemented on Raylib (there is no `System.Drawing.Color` → `Raylib_cs.Color` converter yet; NineSlice's identical gap is marked `// todo - need to convert`). This preserves current (non-)behavior; no regression.
- The `AtlasedTexture` check in `AssignSourceFileOnSprite` is `#if !RAYLIB`-only — `AtlasedTexture` isn't compiled into `RaylibGum.csproj` at all. Preserves current behavior.
- `Blend`: the propertyName dispatch is now shared, but the two platforms use genuinely different underlying storage — `#if RAYLIB` assigns Raylib's native `Gum.RenderingLibrary.Blend? Blend` property directly; `#if !RAYLIB` keeps the existing `.ToBlendState()` → XNA `BlendState` conversion.

### Latent bugs found & fixed (regression-tested)
Restructuring Raylib's commented-out cases into live ones surfaced real gaps — assigning these properties via the state/variable system (`GraphicalUiElement.SetProperty`, as opposed to direct C# property assignment) silently no-op'd on Raylib:
- **`Animate` / `CurrentChainName`**: entirely unwired in Raylib's dispatcher. Added pass-through `Animate`/`CurrentChainName` properties to `Runtimes/RaylibGum/Renderables/Sprite.cs` (mirroring the home renderable) so the shared case body compiles and works identically on both platforms. `CurrentChainName` additionally needed the `UpdateTextureValuesFrom` + `UpdateLayout` follow-up calls that were commented out — without them, `TextureWidth`/`TextureHeight` at the `GraphicalUiElement` layer don't sync when the chain switches this way.
- **`RenderTargetTextureSource`**: the dispatch case was missing entirely from the Raylib copy, even though `SpriteRuntime.RenderTargetTextureSource` is supported on Raylib (`#if XNALIKE || RAYLIB`). Added, mirroring home's `#if !FRB` case.
- **`AssignSourceFileOnSprite` missing-file path**: on Raylib, a failed texture load under `MissingFileBehavior.ConsumeSilently` (the default) silently cleared the texture with no notification. Home already fires `PropertyAssignmentError` in this case (used by the tool's error-indicator system) — Raylib now does too.
- **`Blend` was not actually broken** — it looked disabled (commented out in the dispatcher) but the reflection fallback in `SetPropertyOnRenderable` already found and set Raylib's native `Blend` property correctly via `Convert.ChangeType`. It's now handled explicitly (consistent with the shared switch and slightly clearer), not a behavior change.

### Tests
Added `Tests/RaylibGum.Tests/Runtimes/SpriteSetPropertyTests.cs` — 5 tests covering `SetProperty` for `Blend`, `Animate`, `CurrentChainName` (with two chains so the default chain-0 index doesn't mask the assertion), `RenderTargetTextureSource`, and the `PropertyAssignmentError` missing-file case. Verified red against the pre-fix Raylib source for the 4 genuine bugs (temporarily stashed the Raylib production changes and reran); `Blend` passed even pre-fix as expected (reflection fallback already worked).

- `dotnet test Tests/RaylibGum.Tests/RaylibGum.Tests.csproj` — 483 passed, 0 failed
- `dotnet test MonoGameGum.Tests/MonoGameGum.Tests.csproj` — 1844 passed, 0 failed
- `dotnet build MonoGameGum/KniGum/KniGum.csproj` — 0 errors
- No new compiler warnings from the touched files
- No files added/renamed in a way that needs `GumCoreShared.projitems` changes

**FRB CANARY NOT RUN** — this PR was built in a git worktree (`.claude/worktrees/3615-sprite-family`), and the FRB canary (`dotnet build GumCore/GumCoreXnaPc/GumCore.DesktopGlNet6/GumCore.DesktopGlNet6.csproj`) cannot run from a worktree. **Run it from the primary checkout on this branch before merge.** The touched `Gum/Wireframe/CustomSetPropertyOnRenderable.cs` is part of `GumCoreShared.projitems` (FRB-shared); all new/changed lines in the `#if !RAYLIB` live paths are unchanged from before except cosmetic (type qualification, structural reshuffling), so risk is low, but this hasn't been verified by an actual FRB build.

### DIFF
--- a/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
+++ b/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
@@ -568,95 +568,110 @@ public class CustomSetPropertyOnRenderable
     {
         bool handled = false;
 
-        if (propertyName == "SourceFile")
+        switch (propertyName)
         {
-            var asString = value as String;
-            handled = AssignSourceFileOnSprite(sprite, graphicalUiElement, asString);
-
-        }
-        else if (propertyName == nameof(Sprite.Alpha))
-        {
-            int valueAsInt = (int)value;
-            sprite.Alpha = valueAsInt;
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.Red))
-        {
-            int valueAsInt = (int)value;
-            sprite.Red = valueAsInt;
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.Green))
-        {
-            int valueAsInt = (int)value;
-            sprite.Green = valueAsInt;
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.Blue))
-        {
-            int valueAsInt = (int)value;
-            sprite.Blue = valueAsInt;
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.Color))
-        {
-            if (value is System.Drawing.Color drawingColor)
-            {
-                sprite.Color = drawingColor;
-            }
-            else if (value is Microsoft.Xna.Framework.Color xnaColor)
-            {
-                sprite.Color = xnaColor.ToSystemDrawing();
-
-            }
-            handled = true;
-        }
-
-        else if (propertyName == "Blend")
-        {
-            var valueAsGumBlend = (RenderingLibrary.Blend)value;
-
-            var valueAsXnaBlend = valueAsGumBlend.ToBlendState();
-
-            sprite.BlendState = valueAsXnaBlend;
-
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.Animate))
-        {
-            sprite.Animate = (bool)value;
-            handled = true;
-        }
-        else if (propertyName == nameof(Sprite.CurrentChainName))
-        {
-            sprite.CurrentChainName = (string)value;
-            graphicalUiElement.UpdateTextureValuesFrom(sprite);
-            graphicalUiElement.UpdateLayout();
-            handled = true;
-        }
-#if !FRB
-        else if(propertyName == nameof(SpriteRuntime.RenderTargetTextureSource))
-        {
-            var runtime = graphicalUiElement as SpriteRuntime;
-            if(runtime != null)
-            {
-                if(value == null)
+            case "SourceFile":
                 {
-                    runtime.RenderTargetTextureSource = null;
+                    var asString = value as String;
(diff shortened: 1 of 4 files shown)


vchelaru:
FRB canary run from the primary checkout: `dotnet build GumCore/GumCoreXnaPc/GumCore.DesktopGlNet6/GumCore.DesktopGlNet6.csproj` — **0 errors** (1962 pre-existing warnings, none introduced by this change). Safe to merge.

PR #3696, opened 2026-07-14
Fix case-only folder renames being blocked/broken on Windows

vchelaru:
## Summary
- A Windows user renamed a Screens/Components subfolder to a name differing only by case (e.g. `GameMenuScreens` -> `gamemenuscreens`); on a Linux checkout the project failed to load with "Missing file" for every element in that folder.
- Root cause: `Directory.Exists` (and NTFS in general) is case-insensitive, so `RenameFolderDialogViewModel.OnAffirmative` misdiagnosed the case-only rename target as "already exists" and blocked it. Even bypassing that check, `FileCommands.MoveDirectory`'s manual copy-then-delete-source logic also breaks on a case-only rename (source/destination are the same physical directory, so the final `Directory.Delete(source)` throws "not empty").
- Net effect: on Windows there was no way to actually rename a folder's case through the tool, so the on-disk casing silently drifted from what element names/`.gumx` reference - invisible on Windows/macOS, but breaks the project on a case-sensitive filesystem.

## Fix
- `RenameFolderDialogViewModel.OnAffirmative`: skip the "already exists" guard when the new path is the same directory, differing only by case.
- `FileCommands.MoveDirectory`: when source/destination are the same directory case-insensitively, use `Directory.Move` (which handles case-only renames correctly) instead of the manual merge logic.

## Test plan
- [x] `RenameFolderDialogViewModelTests.OnAffirmative_WhenNewFolderNameDiffersOnlyByCase_ShouldRenameFolderOnDisk` - red before the fix ("Folder ... already exists."), green after.
- [x] `FileCommandsTests.MoveDirectory_WhenSourceAndDestinationDifferOnlyByCase_ShouldRenameDirectoryOnDisk` - red before the fix (`IOException: directory not empty`), green after.
- [x] Full `GumToolUnitTests` suite green (1065 passed).
- [x] `dotnet build GumFull.sln` clean.
- [x] Manual: rename a folder in the Screens/Components tree to a case-only variant of its current name and confirm it succeeds and the folder's on-disk casing actually updates.

### DIFF
--- a/Gum/Commands/FileCommands.cs
+++ b/Gum/Commands/FileCommands.cs
@@ -77,6 +77,20 @@ public class FileCommands : IFileCommands
 
     public void MoveDirectory(string source, string destination)
     {
+        // A rename that only changes casing (e.g. "GameMenuScreens" -> "gamemenuscreens") points
+        // source and destination at the same physical directory on a case-insensitive filesystem
+        // (Windows/macOS). The general merge-into-destination logic below no-ops the "create" and
+        // "move each file into itself" steps, then throws on the final Directory.Delete(source)
+        // because the directory is still non-empty. Directory.Move handles this case correctly.
+        bool isSameDirectoryDifferentCase =
+            string.Equals(NormalizeDirectoryPath(source), NormalizeDirectoryPath(destination), StringComparison.OrdinalIgnoreCase);
+
+        if (isSameDirectoryDifferentCase)
+        {
+            Directory.Move(source, destination);
+            return;
+        }
+
         Directory.CreateDirectory(destination);
 
         // Move files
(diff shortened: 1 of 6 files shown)


vchelaru:
Found and fixed a third bug in the same family while verifying folder-rename cascades correctly to per-element renames: `RenameLogic.RenameXml` saves the element under its new (case-updated) path, then deletes the old path if it still exists. On a case-only rename, old/new are the same physical file (case-insensitive filesystem), so the delete step was destroying the just-saved content - this fires for every screen/component inside a case-renamed folder, and also for a direct case-only rename of a single element's own name. Fixed by using `File.Move` to correct the on-disk casing instead of delete, when old/new differ only by case. Pinned with `RenameLogicTests.HandleRename_WhenElementNameChangesOnlyByFolderCase_ShouldNotDeleteTheJustSavedFile` (red before: file gone after rename; green after).

PR #3697, opened 2026-07-14
Port Cursor event-failure diagnostic to FRB1

vchelaru:
Brings the MonoGame `Cursor.GetEventFailureReason` diagnostic ("why isn't this click registering") to FlatRedBall (FRB1), which has its own `FlatRedBall.Gui.Cursor`.

- `MonoGameGum/Input/CursorExtensions.cs` is now compiled into `FlatRedBall.Forms` via `#if FRB` (paired `.projitems` change in the FlatRedBall repo). MonoGame behavior is unchanged (guarded by `#if !FRB`).
- Diagnostic gains: coordinate-decoupling detection (UI drawn through a zoom / render target the hit-test doesn't see), per-node absolute bounds in the ancestor tree, and cursor / zoom / canvas context.
- Bug fix: `GetInvisibleParent` returned at the parentless root without testing the root's own `Visible`, so an invisible top-level window read as "no failure". Surfaced while debugging a real FRB app whose UI screen toggled invisible per post-processing stage.

Tests: 17/17 in `MonoGameGum.Tests`. FRB Forms canary builds clean. Verified live in an FRB app.

Requires the paired FlatRedBall PR (projitems `<Compile Include>`) to reach FRB users.


vchelaru:
Paired FlatRedBall PR: vchelaru/FlatRedBall#1887 (projitems Include). Merge together.

PR #3805, opened 2026-07-18
Fix mouse back/forward buttons: stop tree-view mis-select, add selection history navigation

### LINKED ISSUE #3804: Mouse Back/Forward buttons don't navigate selection (and wrongly select nodes in the tree view)
**Problem**

Mouse "back"/"forward" side buttons (XButton1/XButton2) don't work anywhere in the tool. Worse, hovering the element tree view and clicking either button *selects* whatever node the cursor happens to be over — an unrelated side effect, not navigation.

Root cause of the tree view mis-select: `MultiSelectTreeView.OnMouseUp` only excludes `MouseButtons.Right` when deciding whether to select the hovered node, so any other button (XButton1/XButton2, Middle) falls through and selects too:

```csharp
var shouldSelect = (EffectiveModifiers() == Keys.None && MultiSelectBehavior != MultiSelectBehavior.RegularClick) &&
    ((mSelectedNodes.Count > 1 && mSelectedNodes.Contains(node)) || IsSelectingOnPush == false) &&
    e.Button != MouseButtons.Right;
```

More broadly, there's no selection-history stack at all, so even once the buttons are wired up correctly there's nothing for them to drive.

**Suggested improvement**

1. Fix `MultiSelectTreeView.OnMouseUp` to require `e.Button == MouseButtons.Left` (allow-list) instead of `!= MouseButtons.Right` (deny-list).
2. Add a selection-history stack with browser-style Back/Forward semantics, recording real selections via the existing `PluginManager.InstanceSelected`/`ReactToStateSaveSelected` cascade, with a re-entrancy guard (mirroring `IsInUiInitiatedSelection`/`SuppressCallAfterClickSelect`) so that navigating Back/Forward doesn't get re-recorded as a new selection and truncate the forward stack.
3. Wire the mouse Back/Forward buttons to drive that stack:
   - XNA wireframe canvas: add `XButton1`/`XButton2` push-detection getters to `Cursor.cs` (MonoGame's `MouseState` already exposes them; the class already polls `Mouse.GetState()` every frame).
   - Tree view: raw `WM_XBUTTONDOWN`/`UP` land directly on the tree view's native child window, so it needs its own handling in `MultiSelectTreeView`.
   - WPF chrome (menus, variable grid, etc.): a `Window`-level `PreviewMouseDown` checking `e.ChangedButton` covers pure-WPF surfaces. It won't see clicks over the tree view or the XNA canvas, since raw XButton messages don't bubble to parent windows the way `WM_APPCOMMAND` does — those two surfaces need the explicit handling above regardless.

**Source**

- `Gum/CommonFormsAndControls/MultiSelectTreeView.cs:247-250` (`OnMouseUp` mis-select)
- `Gum/Plugins/InternalPlugins/TreeView/ElementTreeViewCreator.cs:210,220` (`IsSelectingOnPush = false`, `MultiSelectBehavior = CtrlDown` — why that branch is reachable)
- `InputLibrary/Cursor.cs` (per-frame mouse state polling for the XNA canvas; add XButton getters here)
- `Gum/ToolStates/SelectedState.cs`, `Gum/Plugins/PluginManager.cs` (existing selection cascade + re-entrancy pattern to hook the history stack into)

vchelaru:
Closes #3804

Two related fixes for the mouse back/forward side buttons:

1. **Tree view mis-select bug** - `MultiSelectTreeView.OnMouseUp` only excluded `Right` from selecting the hovered node, so Back/Forward (and Middle) fell through and mis-selected whatever was under the cursor. Fixed to allow-list `Left` only.
2. **Selection history navigation** - new `SelectionHistoryService` (browser-style back/forward stack over element/instance selections), fed by a `SelectionHistoryPlugin` that records real selections. Wired to the mouse buttons on the tree view and, via `MainWindow.PreviewMouseDown`, general WPF chrome (menus, variable grid, dialogs).

**Scoped out (follow-up):** the XNA wireframe canvas isn't wired to the buttons yet.

Also reorders `Gum.sln` so `Gum` is the first project entry - Visual Studio defaults the startup project to the first listing when a worktree has no `.suo` yet, so this makes freshly-opened worktrees default to the right startup project.

## Test plan
- [x] `SelectionHistoryTests` (Gum.Presentation.Tests) - stack semantics: dedup, truncate-on-new-branch, re-entrancy guard
- [x] `MultiSelectTreeViewNavigationTests` (GumToolUnitTests) - mis-select regression + new nav events
- [x] Full `GumFull.sln` build - 0 errors
- [x] `Gum.Presentation.Tests` 259/259, `GumToolUnitTests` 1024/1028 (4 unrelated pre-existing skips), `MonoGameGum.Tests` 1884/1884
- [x] Manual test in progress

### DIFF
--- a/Gum.sln
+++ b/Gum.sln
@@ -3,6 +3,8 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 18
 VisualStudioVersion = 18.0.11012.119
 MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Gum", "Gum\Gum.csproj", "{44575902-06A3-424F-A94C-4B7B8EC3ED2C}"
+EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "XnaAndWinforms", "XnaAndWinforms\XnaAndWinforms.csproj", "{C39A973C-66D6-4A6C-82B5-AE0042F210F8}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "InputLibrary", "InputLibrary\InputLibrary.csproj", "{938D9A00-8529-4CE8-9077-E0A3994A73EA}"
@@ -34,8 +36,6 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{E9BDDA50-D0CD-4172-959F-66E2C1DFEB50}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tools", "Tools", "{A1B2C3D4-E5F6-7890-ABCD-EF1234567890}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Gum", "Gum\Gum.csproj", "{44575902-06A3-424F-A94C-4B7B8EC3ED2C}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "WpfDataUi", "WpfDataUi\WpfDataUi.csproj", "{7261EDF9-108A-4563-8B73-6B72DED72AE9}"
 EndProject
--- a/Gum/CommonFormsAndControls/MultiSelectTreeView.cs
+++ b/Gum/CommonFormsAndControls/MultiSelectTreeView.cs
@@ -120,6 +120,13 @@ namespace CommonFormsAndControls
         /// </summary>
         public event Action<Exception>? UnhandledException;
 
+        /// <summary>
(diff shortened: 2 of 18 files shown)


vchelaru:
Added Alt+Left/Alt+Right hotkeys for the same back/forward navigation (browser convention), dispatched through the existing HotkeyManager app-wide hotkey switch and registered in the Hotkeys panel. Neither key was previously bound in Gum.

vchelaru:
Also fixed the DragDrop/STA crash you kept hitting while I was running the full test suite. Root cause: TreeNode.Expand()/Collapse() and Control.OnMouseDown's base call can both force a WinForms TreeView's native window handle to be created, which requires an STA thread — the test host's isn't guaranteed to be one, so it intermittently crashed the whole process (not something you could catch with try/catch, since it originates inside a native WndProc callback). Fixed by running the affected tests on a dedicated STA thread, mirroring a pattern already used elsewhere in this project for WPF-touching tests. Confirmed clean across two full-suite runs post-fix (1026/1030 passed, 4 unrelated pre-existing skips, 0 failed, no crash).

Issue #3755, opened 2026-07-17
Decouple Gum tool UI from logic — Phase 4a (ElementTreeViewManager / MultiSelectTreeView off TreeNode)


vchelaru:
Tracks Phase 4a of `Direction/ui-decoupling-plan.md`: decouple `ElementTreeViewManager` (~3k lines) and `MultiSelectTreeView` from WinForms `TreeNode`.

**Source of truth:** `Direction/ui-decoupling-plan.md` (Phase 4a). The already-migrated state tree (`StateTreeViewRightClickService`, ex-`StateTreeViewManager`) is the proven template — see #3245/#3246. `Tools/Gum.Presentation/Managers/ITreeNode.cs` already has a neutral tree-node abstraction pattern from that migration to reuse/extend.

## Scope

- [ ] Decouple `ElementTreeViewManager` from `System.Windows.Forms.TreeNode`.
- [ ] Decouple `MultiSelectTreeView` from WinForms tree-view types.

Multi-week; can overlap with Phase 4b (#3756). Not started.

vchelaru:
### Progress ledger — headless `ITreeNode` layer

Building `ITreeNode` (`Tools/Gum.Presentation/`) into a full stand-in so ETVM's `TreeNode`-native internals can convert:

- #3813 — extract icon-index logic into testable `TreeNodeImageLogic`
- #3814 — headless container/folder predicates
- #3815 — element-type predicates
- #3816 — `Children` navigation primitive + `GetAllChildrenNodesRecursively(ITreeNode)`; value equality on `TreeNodeWrapper` (keyed on wrapped node)
- #3817 — mirror ETVM's container-search family (`GetTreeNodeForTag`, `GetTreeNodeFor(InstanceSave)`, `GetInstanceTreeNodeByName`) as `ITreeNode` extensions
- #3819 — mirror ETVM's root-selecting search overloads (`GetTreeNodeFor(ElementSave?/ScreenSave/ComponentSave/StandardElementSave/BehaviorSave)`, `GetTreeNodeForTag` root-fallback) via new `IElementTreeRoots` + `TreeNodeRootSearchExtensions`; ETVM implements `IElementTreeRoots` by wrapping its 4 WinForms root fields in `TreeNodeWrapper`
[...]

PR #3835, opened 2026-07-20
Add ITreeNodeMutable + GumTreeNode: zero-allocation mutation node for ETVM's hot path

### LINKED ISSUE #3832: Decouple Gum tool UI from logic — Phase 4a continued: convert ElementTreeViewManager/MultiSelectTreeView off native TreeNode
## Problem

#3755 built a headless `ITreeNode` mirror API (`Tools/Gum.Presentation/Managers/ITreeNode.cs` + `TreeNode*Extensions.cs` — search/predicate/navigation) and extracted `MultiSelectTreeView`'s testable decision logic. Neither `ElementTreeViewManager` nor `MultiSelectTreeView` has actually moved off native WinForms `TreeNode` — that's still the real Phase 4a goal.

## What's needed

- **ElementTreeViewManager**: construction/mutation/refresh/drag-drop internals still run on `TreeNode` directly. Rewiring to the existing `ITreeNode` mirror as-is regresses perf (wrapper-per-node allocation on the per-instance refresh hot path). Needs either a mutation-capable headless API or an accepted perf tradeoff.
- **MultiSelectTreeView**: still a `System.Windows.Forms.TreeView` subclass. Replacing it is a structural UI change, not a logic extraction — the already-migrated state tree (referenced in `Direction/ui-decoupling-plan.md`) is the proven template.

## Source

#3755 (closed) has the extraction history and the mirror-layer build-out.

vchelaru:
Part of #3832 (Phase 4a "shrink first, swap later"). Introduces the mutation-capable node abstraction ETVM's construction/refresh hot path was missing, without touching that hot path's logic yet.

## What
- `ITreeNodeMutable : ITreeNode` (`Tools/Gum.Presentation/Managers/ITreeNodeMutable.cs`) - adds `ImageIndex`, `SetTag`, and child add/insert/remove/clear to the existing read-only `ITreeNode`.
- `GumTreeNode : TreeNode, ITreeNodeMutable` (`Gum/Plugins/InternalPlugins/TreeView/GumTreeNode.cs`) - implements the interface directly on the concrete WinForms node instead of via a per-call wrapper (`TreeNodeWrapper`'s approach), so nothing allocates to use it. `System.Windows.Forms.TreeView`/`MultiSelectTreeView` don't care what `TreeNode` subtype is in their `Nodes` collection, so this drops in with zero changes elsewhere.
- Swapped ElementTreeViewManager's 9 node-construction call sites (`new TreeNode()` / `.Nodes.Add(string)`) to `GumTreeNode`. Every other `TreeNode`-typed line in the file is untouched.
- `ITreeNodeMutable`/`GumTreeNode` are **not wired into any ETVM logic yet** (`RefreshUi`, `SortByName`, removal) - this PR only lands the enabling type; converting call sites is follow-up work.

## Why this shape
318 `TreeNode` refs in ETVM.cs split into two very different traffic profiles: low-frequency/event-driven (selection, drag-drop, hotkeys) already migrated to the headless `ITreeNode` mirror via `TreeNodeWrapper` in prior work (#3755 and its follow-ups); high-frequency (`RefreshUi` and friends, called on every tree edit, walking full subtrees) never had a mutation-capable option, and wrapping every visited node in a new object per refresh was the reason a naive `ITreeNode`-widening was rejected. Making the concrete node implement the interface (rather than wrapping it) removes that cost entirely.

MultiSelectTreeView needs no changes for this step - it never constructs `TreeNode`s itself (verified: no `new TreeNode(`/`.Nodes.Add` in that file), it only consumes nodes ETVM built. The "these two files must move together" framing from the prior scoping only applies to the eventual widget swap, not to this node-abstraction step.

## Testing
- `GumTreeNodeTests` (12 tests) pin `GumTreeNode`'s dual nature: WinForms `TreeNode` API and `ITreeNodeMutable` API operate on the same underlying storage, `Parent`/`Children` return the same instance (no wrapper) when neighbors are also `GumTreeNode`, and fall back to `TreeNodeWrapper` for a plain `TreeNode` neighbor.
- Full `GumToolUnitTests` suite: 1078 passed, 4 pre-existing skips, 0 failures.

## Manual test verdict
Needed - this changes the concrete type of every node in the live tool's tree view, even though it's designed to be behavior-preserving (GumTreeNode inherits everything from TreeNode; nothing calls the new interface members yet).

Steps once VS is open:
1. Load any project, confirm Screens/Components/Standard/Behaviors tree renders and icons look normal.
2. Add/rename/delete an element and an instance; confirm the tree updates as before.
3. Drag-drop an element into a folder; confirm it still works.
4. Undo/redo a tree-affecting change; confirm the tree reflects it.

## Next step for follow-up work
Start converting `RefreshUi`/`AddTreeNodeForInstance`/`SortByName`'s internals to operate through `ITreeNodeMutable` where it doesn't require touching call signatures outside ETVM - now that the enabling type exists and costs nothing to use.

Closes: not claimed - this does not complete #3832's off-`TreeNode` goal.

### DIFF
--- a/Gum/CommonFormsAndControls/MultiSelectTreeView.Theming.cs
+++ b/Gum/CommonFormsAndControls/MultiSelectTreeView.Theming.cs
@@ -1248,12 +1248,32 @@ public partial class MultiSelectTreeView
 
     #endregion
 
-    private static TreeNode[] ExtractDraggedNodes(IDataObject? data)
+    // Deliberately does not use GetDataPresent(typeof(TreeNode))/GetData(typeof(TreeNode)): WinForms'
+    // default DataObject stores an untyped payload under a format keyed by the payload's *runtime*
+    // type full name (Control.DoDragDrop wraps a single dragged item in `new DataObject(item)` with
+    // no explicit format), not its static/base type. A TreeNode subclass (e.g. GumTreeNode) is then
+    // stored under its own type name and an exact `typeof(TreeNode)` format lookup misses it entirely.
+    // Scanning the actual formats present and pattern-matching on TreeNode/TreeNode[] is robust to any
+    // subclass, present or future, regardless of how the sender packed it.
+    internal static TreeNode[] ExtractDraggedNodes(IDataObject? data)
     {
-        if (data?.GetDataPresent(typeof(TreeNode[])) == true)
-            return (TreeNode[])data.GetData(typeof(TreeNode[]))!;
-        if (data?.GetDataPresent(typeof(TreeNode)) == true)
-            return new[] { (TreeNode)data.GetData(typeof(TreeNode))! };
+        if (data == null)
+        {
+            return Array.Empty<TreeNode>();
+        }
+
+        foreach (string format in data.GetFormats())
+        {
+            if (data.GetData(format) is TreeNode[] nodes)
+            {
+                return nodes;
+            }
+            if (data.GetData(format) is TreeNode node)
+            {
+                return new[] { node };
+            }
+        }
+
         return Array.Empty<TreeNode>();
     }
 
--- a/Gum/CommonFormsAndControls/Properties/AssemblyInfo.cs
+++ b/Gum/CommonFormsAndControls/Properties/AssemblyInfo.cs
@@ -1,5 +1,8 @@
 ﻿using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+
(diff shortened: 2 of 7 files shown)


vchelaru:
Manual test caught a real drag-drop regression pre-merge — fixed and pushed. See the commit message for details; `ExtractDraggedNodes` no longer does an exact-type format match, so it's robust to any TreeNode subclass. Full suite green (1081 passed).

PR #3873, opened 2026-07-21
Relocate ProjectManager's load-repair logic + fix WinForms startup blocker, part of #3863

### LINKED ISSUE #3863: Move ProjectManager's project-management logic into headless Gum.Presentation (Avalonia north-star)
## Problem

Per `Direction/ui-decoupling-plan.md`'s north-star test: could an Avalonia UI reuse Gum's project-management logic today, with only a new View? Not yet checked. `Gum/Managers/ProjectManager.cs` (~1006 lines) sits in `Gum.csproj` (WPF), already constructor-injected (`IProjectManager, IDeleteProjectProvider, ICopyPasteProjectProvider, IReferenceFinderProjectProvider, IRenameProjectProvider`), and has only 2 `System.Windows` text hits — but a low grep count doesn't mean WPF-free (check per-member for WPF-typed parameters/dependencies and dialog calls, not just the namespace text).

## What's needed

Scope (don't implement blind): read the file fully and classify its logic by (a) genuinely relocatable now into `Gum.Presentation`, (b) convertible with reasonable effort, (c) genuinely stuck on real WPF/WinForms glue (dialogs, etc.) — document *why* for (c). Implement a bounded first slice only if one is clearly safe and TDD-backed; otherwise report the categorization and stop.

## Source

`Gum/Managers/ProjectManager.cs`. Related: `Direction/ui-decoupling-plan.md`'s north-star section and "Known gotchas" list.

vchelaru:
## Summary

Scoping + a bounded first slice for #3863 (`Gum/Managers/ProjectManager.cs`, ~1006 lines).

**Implemented (this PR):**
- Extracted the four load-time `GumProjectSave` repair passes (`RemoveSpacesInVariables`,
  `FixRecursiveAssignments`, `FixSlashesInNames`, `RemoveDuplicateVariables`) into a new
  `IGumProjectRepairLogic`/`GumProjectRepairLogic` in `Gum.Presentation` — they already only took a
  `GumProjectSave` parameter, so they carried no WPF/WinForms dependency to begin with.
- Fixed a real WinForms blocker: `Initialize()`'s startup shift-check read `Control.ModifierKeys`
  directly. Converted to `IHotkeyManager.IsPressedInControl(KeyCombination.Shift())` — the same
  live-modifier-state seam already used by `SelectionManager.IsShiftDown`. Verified logically
  identical (both reduce to `(Control.ModifierKeys & Keys.Shift) == Keys.Shift`, and `Keys.Shift`
  is a single-bit flag).

**Scope classification for the rest of `ProjectManager`:**
- **(a) Genuinely relocatable now:** the four extracted methods (done), plus `MakeAbsoluteIfNecessary`
  (already on `IProjectManager`, pure `FileManager` string logic — no blocker at all), `CopyLinkedComponents`
  and `RecreateMissingStandardElements` (both only need `_gumProjectSave` passed as a parameter instead
  of read off the field — a mechanical, not architectural, change; `RecreateMissingStandardElements`
  already has 2 pinning tests).
- **(b) Convertible with reasonable effort:**
  - `WpfDataUi.Controls.FilePickingLogic.FolderRelativeTo` (2 call sites, `LoadProject`/
    `AskUserForProjectNameIfNecessary`) — real WPF (`WpfDataUi.csproj` is `UseWPF`/`UseWindowsForms`).
    Needs a narrow interface exposing just this folder-relative-to setter, WPF-side impl, same shape
    as the `IAppScaleProvider` gotcha in the decoupling plan.
  - `ICommandLineManager` and `IRetryService` — both interfaces are physically declared inside
    `Gum.csproj`-side files, not `Gum.Presentation`. Their own dependency closures are clean
    (`IRetryService` is trivial; `ICommandLineManager` only depends on other already-Presentation
    interfaces plus `IFontManager`, which itself lives in `Gum.csproj` too and needs the same
    physical-file move). Per the "interface relocation is the real gate" gotcha, `ProjectManager`
    can't move to `Gum.Presentation` until these interface files physically relocate — pure file
    moves, no logic change, but out of this PR's bounded scope.
- **(c) Genuinely stuck (real blockers, no clean seam yet):**
  - The `GeneralSettingsFile` property is typed as the concrete `Gum.Settings.GeneralSettingsFile`,
    which itself has WinForms-typed members (`Rectangle MainWindowBounds`, `FormWindowState
    MainWindowState`) — already documented in the class's own doc comment as the reason it's kept
    out of `IProjectManager`'s headless interface surface. This blocks the **concrete `ProjectManager`
    class** from moving even once every other blocker above is fixed, since a public member can't be
    typed against a WinForms type from a headless assembly. Splitting `GeneralSettingsFile`'s
    WinForms-only members off (or `#if`-gating them) is its own design decision — flagging, not
    implementing.

**Net:** the file is not close to fully headless-movable as a whole class yet — this PR advances the
*logic* (repair passes + the shift-check) but the *host class* (`ProjectManager` itself) still has 3
real, distinct blockers (`WpfDataUi`, `ICommandLineManager`/`IRetryService` interface location,
`GeneralSettingsFile`). Leaving #3863 open with this status; did not include `Closes`.

## Test plan
- [x] `dotnet build Tools/Gum.Presentation/Gum.Presentation.csproj` — 0 errors
- [x] `dotnet build GumFull.sln` — 0 errors
- [x] `dotnet test Tests/Gum.Presentation.Tests/Gum.Presentation.Tests.csproj --filter GumProjectRepairLogicTests` — 8/8 pass (new pinning tests)
- [x] `dotnet test Tool/Tests/GumToolUnitTests/GumToolUnitTests.csproj --filter ProjectManagerTests` (with `-p:SolutionDir`) — 10/10 pass (8 existing + 2 new)
- [x] Mutation-test proof: broke `FixRecursiveAssignments`' `Container` assignment, confirmed
      `FixRecursiveAssignments_ForcesRecursiveInstanceToContainer` went red, reverted, confirmed green.

**Manual test: not needed** — behavior-preserving extraction + one seam swap verified logically
identical to the original WinForms read, both covered by unit tests + compilation.

**FRB canary: not needed** — no FRB1-shared source touched (`ProjectManager.cs`/`GumProjectRepairLogic.cs`
aren't referenced by any `.projitems`).

### DIFF
--- a/Gum/Managers/ProjectManager.cs
+++ b/Gum/Managers/ProjectManager.cs
@@ -2,7 +2,6 @@ using CommunityToolkit.Mvvm.Messaging;
 using Gum.CommandLine;
 using Gum.Commands;
 using Gum.DataTypes;
-using Gum.DataTypes.Variables;
 using Gum.Extensions;
 using Gum.Logic;
 using Gum.Logic.FileWatch;
@@ -24,8 +23,6 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
-using System.Windows;
-using System.Windows.Forms;
 using ToolsUtilities;
 
 namespace Gum;
@@ -51,6 +48,8 @@ public class ProjectManager : IProjectManager, IDeleteProjectProvider, ICopyPast
     // rebuild-fonts path reads GumProjectSave) — deferring it breaks the construction cycle.
     private readonly Lazy<ICommandLineManager> _commandLineManager;
     private readonly IPluginManager _pluginManager;
+    private readonly IHotkeyManager _hotkeyManager;
+    private readonly IGumProjectRepairLogic _gumProjectRepairLogic;
 
     #endregion
 
@@ -124,7 +123,9 @@ public class ProjectManager : IProjectManager, IDeleteProjectProvider, ICopyPast
         IStandardElementsManagerGumTool standardElementsManagerGumTool,
         IRetryService retryService,
         Lazy<ICommandLineManager> commandLineManager,
-        IPluginManager pluginManager)
+        IPluginManager pluginManager,
+        IHotkeyManager hotkeyManager,
+        IGumProjectRepairLogic gumProjectRepairLogic)
     {
         _selectedState = selectedState;
         _elementCommands = elementCommands;
@@ -137,6 +138,8 @@ public class ProjectManager : IProjectManager, IDeleteProjectProvider, ICopyPast
         _retryService = retryService;
         _commandLineManager = commandLineManager;
         _pluginManager = pluginManager;
+        _hotkeyManager = hotkeyManager;
+        _gumProjectRepairLogic = gumProjectRepairLogic;
         // Default settings until LoadSettings() replaces this with the loaded (or newly-created)
         // file. Avoids a null-before-load window that every narrowed IProjectManager member above
         // would otherwise need to guard against (some plugins read these before LoadSettings runs -
@@ -155,7 +158,7 @@ public class ProjectManager : IProjectManager, IDeleteProjectProvider, ICopyPast
 
         if (!_commandLineManager.Value.ShouldExitImmediately)
         {
-            var isShift = (Control.ModifierKeys & Keys.Shift) != 0;
+            var isShift = _hotkeyManager.IsPressedInControl(KeyCombination.Shift());
 
             if (!isShift && !string.IsNullOrEmpty(_commandLineManager.Value.GlueProjectToLoad))
             {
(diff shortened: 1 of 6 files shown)


vchelaru:
Fixed a construction cycle introduced by the initial commit: `IHotkeyManager` was injected non-lazily, but `HotkeyManager` depends (via `ISetVariableLogic`) back on `IProjectManager`. Mocked-dependency unit tests never exercise the real DI container, so this only surfaced in CI's `ServiceProviderCompositionSpikeTests` (which resolves the real container) — that's why `Build-Tool` hung for over an hour on the first push. Fixed via `Lazy<IHotkeyManager>`, same pattern as the existing `Lazy<ICommandLineManager>`. All checks green now.

Issue #1934, opened 2026-01-03
Do an allocation pass for Gum


vchelaru:
Test allocations when Gum is:

* Idle
* Reacting to bound property changes
* Scrolling or adjusting layouts

vchelaru:
# Allocation pass — current state & handoff

*Read this comment to pick the work up cold. Goal of this issue: drive **managed heap allocations to zero per frame** in Gum's steady-state runtime loops (idle, bound-property changes, scroll/layout).*

## TL;DR

- **Measurement infrastructure is built** — PR #3513 (branch `1934-alloc-harness`), associated with this issue, **does not close it**.
- **Finding so far:** the layout engine's steady-state relayout already allocates **0 bytes/frame**. The remaining (and likelier) targets are the **Draw walk (idle)**, **binding**, and **Text/Forms**.
- Next step: add the **idle/Draw** scenario (see below).

## What exists (PR #3513, branch `1934-alloc-harness`)

| File | Purpose |
|---|---|
| `Tests/MonoGameGum.TestsCommon/AllocationMeasurer.cs` | The harness. `AllocationMeasurer.Measure(Action, warmupIterations, measuredIterations)` → `AllocationResult { TotalBytes, Iterations, BytesPerIteration }`. Wraps `GC.GetAllocatedBytesForCurrentThread()`. Lives in the shared test lib so any test project can use it. |
| `MonoGameGum.Tests/Performance/AllocationMeasurerTests.cs` | Self-tests: proves the measurer reports exactly `0` for a no-op and `≥` the allocated size for a known `byte[10_000]`. |
| `MonoGameGum.Tests/Performance/LayoutAllocationTests.cs` | Two layout scenarios, both guarding **0 bytes/frame**. |

Run: `dotnet test MonoGameGum.Tests/MonoGameGum.Tests.csproj --filter "FullyQualifiedName~MonoGameGum.Tests.Performance"` → 4/4 green. Add `--logger "console;verbosity=detailed"` to see each scenario's printed bytes/frame.

## How to add a new scenario

1. Build a representative scene; run one pass so it starts laid-out.
2. `AllocationResult result = AllocationMeasurer.Measure(() => oneFrameOfWork(), warmupIterations: 50, measuredIterations: 500);`
3. Assert `result.TotalBytes.ShouldBe(0)` — or a budget you then ratchet down as sources are removed.
[...]

PR #3981, opened 2026-07-22
Pool per-realization width buffer in BitmapFont; remove dead ImageData render path (#1934)

vchelaru:
Part of #1934 (runtime allocation pass).

## What
Font/string realization on the Text **render-target** path allocated a fresh `List<int>` for per-line widths every time a Text rebuilt its render target — in `BitmapFont.RenderToTexture2D` and the obsolete `RenderAtlasedTextureToScreen`. Both now reuse a single static buffer, matching the existing static `letterInfos` pattern (rendering is single-threaded and non-reentrant).

Also removed dead code found while in the area: the unreferenced private `RenderToTexture2DUsingImageData` method and its exclusively-used **16 MB** static `mColorBuffer` (a `2048x2048` `Color[]`), plus the now-unused `using System.Collections;`.

## Allocation measurement
Isolated render-target realization of an unchanged single-line Text (per-frame render-state recording reset each iteration so the measurement reflects only the Text/BitmapFont work, not SpriteBatch machinery):

| | bytes/realization |
|---|---|
| before | 72 |
| after | **0** |

The character-by-character render path was already allocation-free (static `widths` + static `letterInfos` from earlier PRs); this covers the render-target realization those didn't.

## Test
New `FontRealizationAllocationTests` (integration project — realization needs a live `GraphicsDevice`). Red at 72 B/realization before the fix, green at 0 after. Liveness guard proves realization actually ran each iteration (the refresh flag is consumed true→false and a texture is produced).

## Verification
- `MonoGameGum.Tests`: 1892 passed
- Text/render/font/allocation integration tests: 46 passed
- KniGum / RaylibGum / SkiaGum build clean (no new warnings)

Behavior-preserving.

### DIFF
--- a/RenderingLibrary/Graphics/Fonts/BitmapFont.cs
+++ b/RenderingLibrary/Graphics/Fonts/BitmapFont.cs
@@ -3,7 +3,6 @@ using Microsoft.Xna.Framework.Graphics;
 using RenderingLibrary.Math;
 using RenderingLibrary.Math.Geometry;
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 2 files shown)


vchelaru:
FRB canary: **pass** — built `GumCore.DesktopGlNet6` clean (0 errors) from the primary checkout with this branch's `BitmapFont.cs`. The deleted `RenderToTexture2DUsingImageData` isn't referenced from FRB1.

PR #4017, opened 2026-07-24
Variable references: resolve AbsoluteX/Y/Left/Top/Right/Bottom/Width/Height

### LINKED ISSUE #4015: Allow Variable References to read Absolute X/Y/Left/Top/Right/Bottom/Width/Height
## Problem

Variable References currently resolve their right-hand side purely by looking up authored `StateSave` variables by name (`RecursiveVariableFinder` / `EvaluatedSyntax`) — there's no way to reference an object's true resolved/absolute size or position regardless of its own units. `Width = Root.Width * 0.6` gives you whatever raw number `Root.Width` is authored as (e.g. a percentage number if `WidthUnits=PercentageOfParent`), not the resolved pixel value.

This blocks modeling CSS `vw`/`vh`-style relationships ("this is 60% of some ancestor's *true* pixel size, regardless of how that ancestor's own size is expressed") purely through authoring, which came up while investigating why HtmlToGum can't produce elements that reflow correctly when the Gum canvas is resized.

## Proposed change

Teach the Variable Reference evaluator (`EvaluatedSyntax` / `VariableReferenceLogic`, `Gum/Plugins/InternalPlugins/VariableGrid/`, plus the underlying `RecursiveVariableFinder` resolution path) to recognize these as valid right-hand-side identifiers, resolving to the runtime's computed value rather than a raw stored variable:

- `AbsoluteX`, `AbsoluteY`
- `AbsoluteLeft`, `AbsoluteTop`, `AbsoluteRight`, `AbsoluteBottom`
- `AbsoluteWidth`, `AbsoluteHeight`

## Explicitly out of scope

Live re-evaluation on resize. Variable references are author-time snapshots — applying one writes a hard value into the state, and nothing keeps it fresh afterward unless something explicitly re-runs `ApplyVariableReferences`/`ApplyAllVariableReferences`. Keeping a reference to `Root.AbsoluteWidth` fresh across a live canvas resize (design-time or runtime) is separate follow-up work, not part of this issue.

## Depends on

#4014 — needs `AbsoluteWidth`/`AbsoluteHeight` properties in place first. `AbsoluteX`/`AbsoluteY`/`AbsoluteLeft`/`AbsoluteTop`/`AbsoluteRight`/`AbsoluteBottom` already exist as properties today, so this issue could technically proceed for those without #4014, but do the full set together once #4014 lands.

## Source

- `Gum/Plugins/InternalPlugins/VariableGrid/` — `EvaluatedSyntax`, `VariableReferenceLogic`
- `Gum/DataTypes/RecursiveVariableFinder.cs` — base identifier resolution

## Target

August release, after #4014.

vchelaru:
Closes #4015.

Variable references can read Absolute* properties (e.g. `RectangleInstance1.X = RectangleInstance.AbsoluteRight`), resolved live against the wireframe. Two bugs prevented that from working:

**1. Live drag** (earlier commits): `ElementCommands.ModifyVariable` fires every drag tick but resolved references without a live root, so `AbsoluteRight` never materialized and reverted on refresh. Fixed by pushing the dragged value onto the live GUE before resolving and passing the live root through.

**2. On load** (this fix): `WireframeObjectManager.RefreshAll` builds the tree and resolves references while `IsAllLayoutSuspended = true`. Under suspension a `Width`/`X` setter updates the GUE's cached field but its `UpdateLayout` no-ops, so the renderable geometry `AbsoluteWidth`/`AbsoluteRight` read stays stale — `AbsoluteRight` came out as `X + 0` instead of `X + Width`. `RefreshAll` resumes layout and calls `UpdateLayout`, but nothing re-resolved the references. Fix: re-apply variable references after `UpdateLayout`, once geometry has synced.

Also adds the core feature: `EvaluatedSyntax` resolves `AbsoluteX/Y/Left/Top/Right/Bottom/Width/Height` against a live `GraphicalUiElement` root, threaded through the apply/validation paths.

**Tests**
- `WireframeObjectManagerVariableReferenceTests` — drives the real `RefreshAll`; red at 394 (X + stale 0), green at 619 (X + Width). Pins the on-load root cause.
- `ApplyVariableReferencesRuntimeTests` — runtime resolution of Absolute* references.
- `ElementCommandsTests.ModifyVariable_*` — drag/release materialization.

Temporary diagnostic hooks used to trace the root cause have been removed.

### DIFF
--- a/.claude/skills/gum-runtime-variable-references/SKILL.md
+++ b/.claude/skills/gum-runtime-variable-references/SKILL.md
@@ -38,7 +38,7 @@ Gum.Expressions (adds Roslyn) — optional NuGet
 Gum Tool    Game (opt-in)
 ```
 
-The decoupling mechanism is `ElementSaveExtensions.CustomEvaluateExpression` — a static `Func<StateSave, string, string, object>` delegate. When null, falls back to `RecursiveVariableFinder` (simple lookups only). When set by `GumExpressionService.Initialize()`, uses Roslyn for full expression support.
+The decoupling mechanism is `ElementSaveExtensions.CustomEvaluateExpression` — a static `Func<StateSave, string, string, GraphicalUiElement?, object>` delegate. When null, falls back to `RecursiveVariableFinder` (simple lookups only). When set by `GumExpressionService.Initialize()`, uses Roslyn for full expression support. The 4th argument is the live, already-laid-out `GraphicalUiElement` being applied against (when one exists) — it lets a reference resolve the runtime-computed Absolute* properties, which don't exist on authored `StateSave` data at all (see `gum-tool-variable-references` for the resolution mechanism). `ApplyVariableReferences(GraphicalUiElement, StateSave)` supplies its own top-level element automatically; `ApplyVariableReferences(ElementSave, StateSave)` only resolves Absolute* references when called with its optional `liveRoot` argument.
 
 After applying variable references, call `GraphicalUiElement.RefreshStyles()` or
 `GumService.Default.RefreshStyles()` to push the updated values to live visuals. For a deep
--- a/.claude/skills/gum-tool-variable-references/SKILL.md
+++ b/.claude/skills/gum-tool-variable-references/SKILL.md
@@ -51,10 +51,10 @@ SetVariableLogic (variable change entry point)
 
 | Class | Location | Role |
 |-------|----------|------|
-| `VariableReferenceLogic` | `Gum/Plugins/InternalPlugins/VariableGrid/` | Tool-side orchestration: validation, reaction to changes, line expansion |
-| `EvaluatedSyntax` | Same directory | Roslyn-based expression parser/evaluator; resolves right-side values via `RecursiveVariableFinder` |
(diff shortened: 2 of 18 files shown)


vchelaru:
Fixed a real bug caught in manual testing: dragging an instance whose sibling references its AbsoluteX/Y/Width/etc. tracked correctly live, but snapped back on release.

Root cause: `ElementCommands.ModifyVariable` (called every drag tick) applied variable references via the `ElementSave` overload with no live root, so it could never resolve an Absolute* reference and never materialized it into `StateSave`. A separate `GraphicalUiElement`-overload call (live-visual-only, never persisted) was the one actually resolving it during the drag — giving the illusion of correct tracking, but nothing durable was written. On release, refreshing from the never-updated `StateSave` scalar snapped it back.

Fix: push the dragged instance's own value onto its live GUE *before* resolving references (not after), and pass the wireframe's `RootGue` as the live root to the `ElementSave`-overload call too. Added a reproducing unit test (`ElementCommandsTests.ModifyVariable_ReferenceReadsAbsoluteRightOfDraggedInstance_MaterializesLiveResolvedValueIntoStateSave`) that failed before the fix and passes after.

Issue #3731, opened 2026-07-16
FRB1 vs MonoGame Runtime property parity audit: Sprite, Text, Container, NineSlice, Polygon, Rectangle, Circle, ColoredRectangle


vchelaru:
## Why

`Gum/Wireframe/CustomSetPropertyOnRenderable.cs` is compiled as FRB1-shared source (via `GumCoreShared.projitems`), so it has to dispatch property-sets for both MonoGame's compiled `Gum.GueDeriving.*Runtime` classes and FRB1's per-game **Glue-generated** runtime equivalents (FRB1 has no compiled Runtime library at all — Glue bakes a `<Type>Runtime.Generated.cs` per standard element directly into each game project). Today that split shows up as scattered `#if FRB` / `#else` pairs, currently concentrated in `TrySetPropertyOnText` (being converged onto `TextRuntime` per [ADR 0009](../blob/main/Direction/decisions/0009-converge-text-dispatch-onto-textruntime.md)).

Direction under discussion: define narrow role interfaces (e.g. `ISpriteRuntime`) in the FRB1-shared source set, have FRB1's Glue-generated runtime and MonoGame's `Gum.GueDeriving.*Runtime` both implement them, and dispatch against the interface instead of the concrete type. That collapses `#if FRB { renderable.X = v } #else { runtime.X = v }` pairs into one unconditional `runtime.X = v` wherever both sides already expose the same property — leaving a single-sided `#if !FRB` (no `#else`, defaulted value) only for genuine capability gaps. Sprite is already being converged on a separate branch/worktree (`3727-sprite-dispatch`).

Before doing this per-type, we want a full property inventory per standard element: what matches, what's MonoGame-only, what's FRB-only, and — separately — what's already unified because it lives directly on the shared `GraphicalUiElement.cs` (no gap possible, since both worlds compile that exact file). That inventory is this issue.

## Method

- **MonoGame/tool side**: `Gum/Managers/StandardElementsManager.cs` — the canonical default-variable list per standard element type (`mDefaults.Add("X", stateSave)` blocks). This is more authoritative than any single `*Runtime.cs`, since it's what the tool serializes/deserializes and what `CustomSetPropertyOnRenderable` ultimately dispatches against.
[...]

PR #4036, opened 2026-07-26
Declare FRB-only runtime interfaces, unify Sprite/NineSlice/Container/Polygon dispatch

vchelaru:
Declares small FRB-only interfaces (`ISpriteRuntime`, `INineSliceRuntime`, `IContainerRuntime`, `IPolygonRuntime`) inside `#if FRB` in `CustomSetPropertyOnRenderable.cs`, matching the property subset Glue's generated runtimes are known to implement (property-parity audit, #3731). Collapses each type's matched `#if FRB / #else` dispatch branches into one unconditional path via the same alias-and-cast idiom the other backends already use.

**Collapsed:**
- Sprite: `Alpha`, `Red`, `Green`, `Blue`, `Blend`
- NineSlice: `Blend`, `CustomFrameTextureCoordinateWidth`, `Red`, `Green`, `Blue`, `BorderScale`, `IsTilingMiddleSections`
- Container: `Alpha`, `IsRenderTarget`
- Polygon: `Alpha`, `Red`, `Green`, `Blue`

**Left alone:** NineSlice/Container `Color` (multi-type conversion logic, not confidently identical on FRB), Polygon `Color`, `SourceFile`, `SourceShaderFile`, `CurrentChainName`, `AnimationChains`, `RenderTargetTextureSource`, `Points`. Text is out of scope (separate ADR 0009 effort).

Nothing outside FRB compiles or sees these interfaces; Glue's codegen implementing them is a separate FlatRedBall-repo follow-up.

Part of #3731

### DIFF
--- a/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
+++ b/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
@@ -32,6 +32,23 @@ using Gum.Wireframe;
 using Gum.GueDeriving;
 #endif
 
+// FRB has no compiled XRuntime types - Glue (FlatRedBall's editor) generates a per-project
+// XRuntime.Generated.cs that this shared Gum file can't reference by concrete type. These FRB-only
+// aliases point at small FRB-only interfaces (declared below, under #if FRB) that mirror the subset
+// of properties Glue's generated runtimes are known to implement with matching names/types, so the
+// FRB dispatch branches below can use the same alias-and-cast idiom the other backends already use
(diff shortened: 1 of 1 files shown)


vchelaru:
Marking draft: the FRB dispatch branches now cast to interfaces (ISpriteRuntime etc.) that Glue's generated runtime classes don't implement yet. That cast is a no-op (null) until the FlatRedBall-side codegen change lands, so this silently breaks Alpha/Red/Green/Blue/Blend/etc. property assignment for every FRB game until the paired FlatRedBall PR ships. Holding until both are ready to merge together.

PR #4075, opened 2026-07-29
SkiaGum text: add [OutlineThickness] BBCode tag, unify onto RichTextKit halo

### LINKED ISSUE #4037: SkiaGum text: [outlinethickness] BBCode tag needs a per-run design (split from #3719)
## Problem

Split from #3719 after its `[font]` half landed. SkiaGum's inline BBCode tag support still has no `[outlinethickness=N]` (per-run outline width override).

This one is harder than `[font]` was: Skia's whole-string `OutlineThickness` deliberately avoids RichTextKit's per-run `Style.HaloWidth` (miter-join spiking at small sizes — see the comment above `BuildRunStyle`'s return in `Runtimes/SkiaGum/Renderables/Text.cs`) and instead does a whole-image recolor+dilate pass in `Render()` (`DrawTextWithOutline`). A per-run version has no obvious equivalent of that dilate pass and needs its own design pass before implementation — not scoped here, just tracked.

## Source

- `Runtimes/SkiaGum/Renderables/Text.cs` — `SupportedTags` (intentionally excludes `outlinethickness`, with a comment explaining why), `BuildRunStyle`, `Render`/`DrawTextWithOutline`

vchelaru:
Closes #4037.

Whole-string `OutlineThickness` used to be a bespoke recolor+dilate render pass because RichTextKit's `Style.HaloWidth` had a hardcoded miter join that spiked at acute glyph corners. The vendored round-join patch (#4068) fixed that, so both whole-string and per-run outline now go through `Style.HaloWidth`/`HaloColor` directly, doubled to compensate for the halo being a centered stroke.

- Added `[OutlineThickness=N]` BBCode tag (per-run, SkiaGum only)
- Deleted `DrawTextWithOutline`'s recolor+dilate pass
- Outline color stays whole-Text only, not per-run

Manual test: opened `Samples/SkiaGumWpfSample` with a whole-string + per-run outline demo (`WAVY` text to stress acute corners). Visual confirmation of no spiking is up to whoever reviews the manual step.


vchelaru:
Also added [OutlineThickness=2] to the shared MonoGame/Raylib/SilkNet TextScreen BBCode demo line (Screens/TextScreen.cs) — SkiaGum was the only backend missing it there. Builds clean for MonoGameGumInCode, raylib GumTest, and SilkNetGumSample.

vchelaru:
Wrapping up here. The per-run outline tag and the halo unification are correct and proven on every backend the test suite covers (SkiaGum.Tests + CPU-raster GumService end-to-end). The live-only SilkNetGum rendering gap (halo invisible on its ANGLE/D3D11 GPU canvas specifically) is unrelated to this PR's logic -- pre-existing on the whole-string path too -- and tracked separately in #4077.

PR #4081, opened 2026-07-29
Restore Gum window focus after external font generation

### LINKED ISSUE #4078: Gum window loses foreground focus after bmfont.exe font generation finishes
## Problem

Running Gum through Visual Studio's debugger: create a new project, add Forms components. The font-generation popup appears after picking a destination (expected), but as soon as font generation finishes, the Gum window drops behind Visual Studio instead of staying in the foreground. No breakpoint is hit - it's a pure window-focus/z-order issue. User has to click Gum in the taskbar to bring it back.

## Root cause

`BmFontExeFileGenerator.GenerateFont` launches `bmfont.exe` with `ProcessStartInfo.UseShellExecute = true`. The shell-executed process gets its own top-level window and steals the foreground. When it exits, Windows doesn't automatically restore foreground focus to Gum - it goes to whatever the OS picks next (Visual Studio, since it's the debugger host here). No code in the call chain calls `Activate()`/`SetForegroundWindow()` on Gum's main window afterward.

Pre-existing since the original KernSmith integration (not a recent regression).

## Suggested improvement

After the bmfont.exe process exits, have the WPF-side caller of font generation re-activate Gum's main window (`mainWindow.Activate()`; may need a `SetForegroundWindow`/`AllowSetForegroundWindow` P/Invoke fallback since Windows throttles `Activate()` calls from a non-foreground process).

## Source

- `Tools/Gum.ProjectServices/FontGeneration/BmFontExeFileGenerator.cs:48-55` - the `UseShellExecute = true` launch.
- `BmFontExeFileGenerator` lives in the headless `Gum.ProjectServices` project (no WPF reference) - the actual fix belongs in whichever WPF caller invokes font generation for Add Forms; that call site still needs to be traced.

vchelaru:
## Summary

- New projects default to `FontGeneratorType.KernSmith` (in-process, no external window), so an earlier version of this PR that hooked `BmFontExeFileGenerator`'s process exit never actually fired for the reported repro.
- The real "popup" from the issue is Gum's own `Spinner` window (`Gum/Controls/Spinner.xaml`), shown during any bulk font-generation pass (Add Forms, project load, etc.) regardless of backend. WPF's `Window.Hide()` doesn't restore focus to its owner, so closing it can hand focus to whatever else is in the z-order (Visual Studio, when running Gum under the debugger).
- Fix: `ToolFontGenerationCallbacks.SpinnerHandle.Dispose()` now calls a new `IGuiCommands.ActivateMainWindow()` right after hiding the spinner. Covers both `BmFont` and `KernSmith` since both share the same spinner.
- `GuiCommands.ActivateMainWindow()` uses `Window.Activate()` plus the standard `AttachThreadInput` workaround to force `SetForegroundWindow` to succeed from a background process (`AllowSetForegroundWindow` only helps when called *by* the current foreground process, so it doesn't apply here).

Closes #4078

## Test plan
- [x] Unit tests (`ToolFontGenerationCallbacksTests`) — green
- [x] `GumFull.sln` builds clean
- [ ] Manual: repro steps below

Manual test: needed — real OS focus/z-order effect, can't be asserted headlessly.
1. Open the worktree's `Gum.sln` in Visual Studio, run Gum under the debugger (F5).
2. Create a new project, then use "Add Forms Components" and pick a destination.
3. Once font generation finishes, Gum should stay in the foreground instead of dropping behind Visual Studio.

### DIFF
--- a/Gum/Commands/GuiCommands.cs
+++ b/Gum/Commands/GuiCommands.cs
@@ -137,7 +137,11 @@ public class GuiCommands : IGuiCommands
     /// <inheritdoc/>
     public void ActivateMainWindow()
     {
-        if (Application.Current?.MainWindow is not { } mainWindow)
+        // IsLoaded: font generation can run before the main window has completed its Show()
+        // sequence (e.g. during initial project load at startup) - Activate() throws
+        // InvalidOperationException ("Cannot call DragMove or Activate before a Window is shown")
+        // if called that early.
+        if (Application.Current?.MainWindow is not { IsLoaded: true } mainWindow)
         {
             return;
         }
--- a/Gum/Commands/GuiCommands.cs
+++ b/Gum/Commands/GuiCommands.cs
@@ -142,24 +142,50 @@ public class GuiCommands : IGuiCommands
             return;
         }
 
-        mainWindow.Activate();
-
-        // Windows throttles Activate() calls from a background process - grant this process
-        // permission to steal foreground focus first, then take it directly via Win32.
         // EnsureHandle forces the Win32 HWND to be created if it doesn't exist yet.
         IntPtr windowHandle = new WindowInteropHelper(mainWindow).EnsureHandle();
-        NativeMethods.AllowSetForegroundWindow(NativeMethods.ASFW_ANY);
-        NativeMethods.SetForegroundWindow(windowHandle);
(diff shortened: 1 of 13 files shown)


vchelaru:
Manual test passed: new project -> Add Forms Components -> font generation finishes -> Gum stays in the foreground, no crash.

PR #4083, opened 2026-07-29
Fix TextRuntime.OutlineColor defaulting to transparent, hiding SkiaGum halo

### LINKED ISSUE #4077: SkiaGum OutlineThickness halo doesn't render live on SilkNetGum
`OutlineThickness` (both the pre-existing whole-string property and the new per-run `[OutlineThickness=N]` BBCode tag from #4037/#4075) doesn't render visibly in `Samples/SilkNetGum/SilkNetGumSample`'s live window.

## Verified working, not the cause
- Style/data model: `Style.HaloWidth`/`HaloColor` compute correctly (unit tests).
- Full render pipeline on headless CPU raster: `GumService` -> `TextRuntime` -> `GraphicalUiElement` -> `Text.Render` produces the outline correctly (confirmed with a throwaway pixel-count test: 1159 outline-colored pixels for a 60pt glyph with a 6px outline).
- Stale build/package: `SilkNetGum.csproj` had its own un-patched `Topten.RichTextKit` reference (fixed in #4075) -- confirmed the correct patched DLL builds and loads (byte-size match against the vendored package), confirmed after a full clean rebuild (bin/obj + build-server cleared). Still not visible live.

## Not yet ruled out
- Something specific to SilkNetGum's live GPU-backed (OpenGLES/ANGLE) Skia canvas vs. the CPU-raster `SKSurface` the headless tests/GumService.Draw() calls use.
- Something in SilkNetGum's own per-frame canvas setup (`Samples/SilkNetGum/SilkNetGumSample/Program.cs`) not exercised by a headless `GumService.Draw()` call.

## Repro
Run `Samples/SilkNetGum/SilkNetGumSample`, open the Text screen, look at "I am text with OutlineThickness = 2" (pre-existing, no BBCode involved) or the "outlined" word in the shared BBCode demo line. Neither shows an outline.

vchelaru:
Closes #4077

Not a GPU/ANGLE bug. `TextRuntime`'s `_outlineColor` backing field (Skia-only) had no initializer, defaulting to fully transparent instead of matching `SkiaGum.Text`'s own opaque-black constructor default. `SystemManagers.UpdateFonts` pushes `OutlineColor` down to the contained `Text` renderable unconditionally whenever `UpdateToFontValues` runs, including as a side effect of setting `OutlineThickness` alone. So setting only `OutlineThickness` (the common case, and what every affected demo does) silently clobbered the renderable's real default to transparent, and RichTextKit's `Style.HaloColor != SKColor.Empty` gate skipped the halo draw entirely.

Confirmed by instrumenting the real render path live on SilkNetGumSample's ANGLE/D3D11 GPU-backed canvas (not just headless CPU raster) -- the halo draw call itself was never reached because `Style.HaloColor` was `#00000000`.

Fix: initialize `_outlineColor` to `SKColors.Black`.

Also adds a screenshot-capture mode to SilkNetGumSample (`GUM_SCREENSHOT_PATH`/`GUM_SCREENSHOT_SCREEN` env vars): snapshots the live GPU-backed surface to a PNG and exits. This is what let me verify a GPU-specific rendering bug without a human watching the window, and it's reusable for any future SilkNetGum rendering issue.

## Test plan
- [x] New unit test (`OutlineThickness_SetWithoutOutlineColor_ShouldNotClobberContainedTextOutlineColorToTransparent`) reproduces the bug through `TextRuntime` (not the renderable directly, which is why existing tests missed it), red before the fix, green after.
- [x] `SkiaGum.Tests` full suite green (393/393).
- [x] `MonoGameGum.Tests` TextRuntime tests green (88/88), `KniGum`/`RaylibGum` build clean (change is `#if SKIA`-gated, no other backend affected).
- [x] Manual: `SilkNetGumSample`'s Text screen, before/after screenshots on the real ANGLE/D3D11 GPU-backed surface -- "outlined" (per-run tag), "Shadow and outline", and "I am text with OutlineThickness = 2" all show the black outline now.

Manual test: needed -- VS is open on `Samples/SilkNetGum/SilkNetGumSample.sln`. Run it, go to the Text screen, and confirm the outline shows on "outlined", "Shadow and outline", and "I am text with OutlineThickness = 2".

FRB canary: not needed -- change is `#if SKIA`-gated in a source file also compiled for FRB1, but the touched code path (`_outlineColor` field initializer) is inside the same `#if SKIA` block already, so FRB (XNALIKE) compiles the untouched branch.

### DIFF
--- a/MonoGameGum/GueDeriving/TextRuntime.cs
+++ b/MonoGameGum/GueDeriving/TextRuntime.cs
@@ -414,7 +414,13 @@ public class TextRuntime : InteractiveGue
     }
 
 #if SKIA
-    Color _outlineColor;
+    // Defaults to opaque black (not default(Color), which is fully transparent) to match
+    // SkiaGum.Text's own constructor default -- UpdateFonts pushes this to the contained Text
+    // unconditionally whenever UpdateToFontValues runs, including as a side effect of setting only
+    // OutlineThickness, so an uninitialized transparent default silently clobbered Text's halo
+    // color and made RichTextKit's `Style.HaloColor != SKColor.Empty` gate skip the halo draw
+    // entirely (issue #4077).
+    Color _outlineColor = SkiaSharp.SKColors.Black;
     /// <summary>
     /// The color of the outline drawn when <see cref="OutlineThickness"/> is greater than zero.
     /// Skia-only: MonoGame/Raylib bake the outline into the font atlas at generation time and expose
--- a/Runtimes/SkiaGum/Renderables/Text.cs
+++ b/Runtimes/SkiaGum/Renderables/Text.cs
@@ -1462,7 +1462,12 @@ public class Text : IRenderableIpso, IVisible, IFormsText, ICloneable
             // per-run halo is used directly, which is also what makes per-run [OutlineThickness]
             // possible (issue #4037) -- RichTextKit paints per-run halo natively in one pass.
             HaloWidth = outlineThickness > 0 ? outlineThickness * 2f : 0f,
-            HaloColor = OutlineColor,
+            // RichTextKit's halo gate (`Style.HaloColor != SKColor.Empty`) ignores HaloWidth, and a
+            // Stroke-style SKPaint with StrokeWidth 0 draws as a 1px hairline instead of nothing -- so
+            // passing OutlineColor through unconditionally drew a thin outline on every glyph whenever
+            // OutlineColor was non-transparent, even with no outline requested (issue #4077 follow-up).
+            // Suppress to Empty whenever there's no actual outline thickness.
+            HaloColor = outlineThickness > 0 ? OutlineColor : SKColor.Empty,
         };
 
         return (style, text, xOffset, yOffset);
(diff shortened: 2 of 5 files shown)


vchelaru:
Also fixed `BoldWeight`: same unconditional-push bridge in `SystemManagers.UpdateFonts` recomputed it from `IsBold` (`IsBold ? 1.5f : 1.0f`) instead of forwarding the stored value, so any custom weight outside {1.0, 1.5} got silently reset on the next unrelated font-property change. Added a red-then-green test and forwarded the stored value instead.

vchelaru:
Caught a regression from my first commit here: making `OutlineColor` default to opaque black exposed a separate latent bug -- RichTextKit's halo gate only checks `HaloColor != Empty`, ignoring `HaloWidth`, and a 0-width Stroke paint draws a 1px hairline instead of nothing. So default text (`OutlineThickness=0`) started showing a thin outline on every glyph. Fixed by suppressing `HaloColor` to `Empty` in `BuildRunStyle` whenever `OutlineThickness <= 0`. Added a red-then-green test (`OutlineThickness_Default_ShouldNotEmitHalo`) and re-verified on the live ANGLE/D3D11 surface -- default text is clean again, requested outlines still render.

Issue #3706, opened 2026-07-15
Converge CustomSetPropertyOnRenderable's Text dispatch onto TextRuntime (runtime-type-first, matching the shape dispatcher)


vchelaru:
## Problem
`CustomSetPropertyOnRenderable`'s Text dispatch is renderable-typed at the top level in both copies (`Gum/Wireframe/CustomSetPropertyOnRenderable.cs` and `Runtimes/SkiaGum/CustomSetPropertyOnRenderable.cs`):

```
SetPropertyOnRenderable
  → else if (containedObjectAsIpso is Text asText)   // renderable-type check
      → TrySetPropertyOnText(Text text, GraphicalUiElement gue, ...)   // Text is the FIRST param
```

Every text property (Blend, MaxLettersToShow, ellipsis, BBCode, dropshadow, outline, Font/FontSize/CustomFontFile/UseCustomFont, ...) is gated behind that renderable check — exactly the pattern Direction 0007/0008 already converged away from for the shape dispatcher (`RectangleRuntime`/`CircleRuntime`/`ArcRuntime`), but Text was never covered by either decision.

This surfaced concretely while fixing #3670/#3703: two new dispatch arms (`UseCustomFont`/`CustomFontFile`) had to be added to a renderable-typed method rather than a Runtime-typed one, because that's the only place font properties currently live in this file that also gets the correct string-path suspension semantics (see the `IsAllLayoutSuspended`-vs-`IsLayoutSuspended` comment atop `Gum/Wireframe/CustomSetPropertyOnRenderable.cs::UpdateToFontValues`).

[...]

PR #4087, opened 2026-07-29
Reconcile TrySetPropertyOnText's Text wrapper with TextRuntime.Text's

vchelaru:
Part of #3706. First of two PRs converging Text's dispatch onto TextRuntime (the ADR 0009 step 2 the issue deferred) -- this one removes the behavior gap that made a direct dispatcher-to-TextRuntime redirect unsafe.

## Problem

`TrySetPropertyOnText`'s "Text"/"TextNoTranslate" branch and `TextRuntime.Text`'s setter each independently wrap the same text-assignment logic with their own width-null/UpdateLayout logic -- but the two wrappers differ:

- Dispatcher: ignores `MaxWidth` when nulling wrap width, and always runs an unconditional, full parent-climbing `UpdateLayout()`.
- `TextRuntime.Text`: respects `MaxWidth`, and only climbs to the parent if the measured wrapped size actually changed (and only as far as a parent that depends on it).

This is why converging the dispatcher onto `textRuntime.Text = value` (the direction #3706 wants, matching the shape-dispatcher pattern) previously looked risky -- it would silently swap one wrapper's behavior for the other across every `SetProperty("Text", ...)` call site (state system / project load), not just direct `.Text =` calls.

## What this does

Added characterization tests comparing both entry points across MaxWidth-wrapping, fixed-width/auto-height, and stack-sibling-position scenarios. All three show byte-identical output today -- the divergence is duplicated logic, not a real behavior split (the dispatcher's cruder approach is self-correcting via its own subsequent full layout pass). Reconciled the dispatcher's wrapper to match `TextRuntime.Text`'s (the later, more precise one), removing the duplication and clearing the way for a follow-up PR to make the dispatcher delegate directly to `textRuntime.Text =`/`SetTextNoTranslate` without changing behavior.

## Testing

- New pinning tests in `TextRuntimeTests.cs`, all green.
- Full `MonoGameGum.Tests` (1952), `MonoGameGum.Tests.V2` (132), and `RaylibGum.Tests` (544) suites green -- no regressions.
- FRB canary (`GumCore.DesktopGlNet6.csproj`): 0 errors, same warning baseline as `main`.

Manual test: not needed -- behavior-preserving, proven by the pinning tests above plus the full regression suite.

### DIFF
--- a/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
+++ b/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
@@ -874,12 +874,22 @@ public partial class CustomSetPropertyOnRenderable
 
         if (propertyName == "Text" || propertyName == "TextNoTranslate")
         {
-            if (graphicalUiElement.WidthUnits == DimensionUnitType.RelativeToChildren ||
-                // If height is relative to children, it could be in a stack
-                graphicalUiElement.HeightUnits == DimensionUnitType.RelativeToChildren)
+            // Mirrors TextRuntime.Text's own wrapper (MonoGameGum/GueDeriving/TextRuntime.cs) so both
+            // entry points measure/re-layout identically instead of each doing it slightly differently.
+            var widthBefore = textRenderable.WrappedTextWidth;
+            var heightBefore = textRenderable.WrappedTextHeight;
+
+            if (graphicalUiElement.WidthUnits == DimensionUnitType.RelativeToChildren)
             {
-                // make it have no line wrap width before assignign the text:
-                textRenderable.Width = null;
+                if (graphicalUiElement.MaxWidth == null)
+                {
+                    // make it have no line wrap width before assignign the text:
+                    textRenderable.Width = null;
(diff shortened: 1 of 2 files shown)


vchelaru:
Correction: this PR does NOT remove the double UpdateLayout call I implied in the description/commit message. A new call-count test (`Text_ViaDirectPropertySet_WhenSizeChanges_StillTriggersTwoUpdateLayoutCalls`, verified red against the pre-fix code) shows `.Text =` still triggers 2 UpdateLayout calls after this PR, same as before -- reconciling the two wrappers' *conditions* made them agree on whether to update, but they're still two separate invocations (TextRuntime.Text's own wrapper, plus the dispatcher's, reached via the SetProperty call in between). The actual call-count reduction only happens in the follow-up PR once the dispatcher delegates directly to `textRuntime.Text =` instead of running its own copy of the logic. This PR's real scope is: make the two wrappers' logic identical (removing the risk they silently diverge), not reduce call count.

PR #4106, opened 2026-07-29
Guard reflection-based runtime type registration against AOT metadata trimming

### LINKED ISSUE #4105: GumService.RegisterRuntimeTypesThroughReflection scans every loaded assembly, breaks under Native AOT
## Problem

Publishing a MonoGame/KNI game as Native AOT throws on `GumService.Initialize`:

```
Unhandled Exception: System.TypeLoadException: The type 'System.Runtime.Intrinsics.X86.Sse3' cannot be found in assembly 'System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
   at System.Reflection.Runtime.General.TypeResolver.Resolve(QTypeDefRefOrSpec, TypeContext)
   at System.Reflection.Runtime.TypeInfos.RuntimeTypeInfo.get_BaseTypeWithoutTheGenericParameterQuirk()
   at System.Reflection.Runtime.TypeInfos.RuntimeTypeInfo.get_BaseType()
   ...
   at Gum.GumService.RegisterRuntimeTypesThroughReflection()
   at Gum.GumService.InitializeInternal(Game, GraphicsDevice, String, SystemManagers, DefaultVisualsVersion)
```

Root cause: `RegisterRuntimeTypesThroughReflection` iterates `AppDomain.CurrentDomain.GetAssemblies()` and calls `assembly.GetTypes()` / `type.GetMethod(...)` on every type in every loaded assembly, including framework assemblies like `System.Private.CoreLib`. Under NativeAOT, unused framework type metadata (base types, generic instantiations, etc.) is trimmed away, so walking all BCL types via reflection hits missing metadata and throws.

## Suggested improvement

Scope the scan down instead of walking every loaded assembly. Options:
- Only scan the entry assembly (already handled separately for the legacy singular `RegisterRuntimeType` hook) plus assemblies that are known Gum extension packages, instead of `AppDomain.CurrentDomain.GetAssemblies()` in full.
- Skip framework/system assemblies explicitly (e.g. filter by assembly name prefix) before calling `GetTypes()`.
- Gate the whole reflection-based fallback behind a check for whether `[ModuleInitializer]`-based registration already ran, since the fallback only exists for WASM/Blazor (module initializers don't fire until a type is touched) and legacy projects.

## Source

`MonoGameGum/GumService.XnaLike.cs:214-249` (`RegisterRuntimeTypesThroughReflection`), called from `InitializeInternal` at line 181.

vchelaru:
Closes #4105

`RegisterRuntimeTypesThroughReflection` scanned every loaded assembly and called `Type.GetMethod` on every type, including framework/BCL assemblies. Under Native AOT, `GetMethod` itself (not just `Invoke`) can throw while resolving trimmed base-type metadata, crashing `GumService.Initialize` entirely.

Fix: skip framework/BCL assemblies by name before scanning (they never define a Gum registration hook), and wrap the `GetMethod`+`Invoke` lookup in one try/catch so a single bad type (ambiguous overload, trimmed metadata) can't abort the whole scan.

Test: added `GumServiceRuntimeTypeReflectionScanTests` — pins the fix using an ordinary JIT-reproducible trigger (an ambiguous `RegisterRuntimeTypes` overload set, which throws `AmbiguousMatchException` from `GetMethod` the same way trimmed AOT metadata throws `TypeLoadException`), plus a table-driven test of the framework-assembly filter.

Manual test: not needed — the failure only reproduces under an actual Native AOT publish, which isn't practical to drive here. Verified instead with the regression-shaped unit test above (red before the fix, green after), the full `MonoGameGum.Tests` suite (1964 passed), the real `Initialize(Game, ...)` path via `MonoGameGum.IntegrationTests`, and a clean `KniGum` build (shares this XNALIKE source).

FRB canary: not needed — `GumService.XnaLike.cs` isn't part of `GumCoreShared.projitems`.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -46,6 +46,78 @@ jobs:
           reporter: dotnet-trx
           fail-on-error: true
 
+  # End-to-end Native AOT smoke test for issue #4105: RegisterRuntimeTypesThroughReflection walked
+  # every loaded assembly via unguarded reflection, which crashed GumService.Initialize under
+  # Native AOT when it hit trimmed framework metadata - a failure mode only a real `dotnet publish
+  # -p:PublishAot=true` reproduces, `dotnet test` always runs JIT and can't catch it. This job
+  # publishes Tests/NativeAotSmoke.Harness as Native AOT and runs it end-to-end (real
+  # GraphicsDevice, real Texture2D upload, default-visuals construction), so a future regression
+  # anywhere in that path is caught automatically instead of waiting for another user report.
+  #
+  # The runner has no GPU, so the harness is forced onto Mesa's llvmpipe software GL the same way
+  # "Install Mesa llvmpipe" does for RaylibGum.Tests below - confirmed to work for MonoGame
+  # DesktopGL too, not just raylib (both resolve opengl32.dll by name from the exe's own directory
+  # before System32).
+  #
+  # Needs neither FNA/Sokol submodules nor mobile workloads - the harness only references
+  # MonoGameGum/GumCommon - so ExcludeAndroid/ExcludeIOS keep restore on just net8.0, same escape
+  # hatches Build-Generated-Code below uses for the same reason.
+  Native-AOT-Smoke:
+    if: github.event_name != 'push'
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: false
+
+      - name: Setup .NET SDKs
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 9.0.x
+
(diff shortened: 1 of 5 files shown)


vchelaru:
Added a permanent regression guard: `Native-AOT-Smoke` CI job (new) publishes `Tests/NativeAotSmoke.Harness` with `PublishAot=true` and runs Gum's real `Initialize` pipeline — real GraphicsDevice (Mesa llvmpipe software GL, no GPU on the runner), real Texture2D upload, default-visuals construction — end to end. Verified it fails with the exact #4105 stack trace against the pre-fix source and passes against the fix; confirmed green on real GitHub Actions just now (not just locally).

PR #4132, opened 2026-07-30
Spatial (any-angle) gamepad/stick focus navigation

### LINKED ISSUE #4129: Spatial (any-angle) focus navigation for gamepad/stick on scattered UI layouts
## Goal

Real, integrated spatial (any-angle) focus navigation for gamepad/stick input, for scattered
(non-grid, non-list) UI layouts — skill trees, radial menus, HUDs — where "nearest in the direction
I'm pointing" is the right model, unlike linear layouts where tab order already works.

## Status

The spike (PR #4132) proved the approach feels right: `SpatialNavigationService` (distance +
angular-alignment scoring) and a first-pass `FrameworkElement.HandleGamepadSpatialNavigation`
method, manually wired in a scratch demo. That validated the algorithm and caught real bugs
(diagonal D-pad detection, an integration hazard with the existing `GamePadsForUiControl`-driven
navigation) — see PR #4132 for details.

**What's not done yet**: real integration into the actual tabbing system. Right now a control has
to manually call a separate method and carefully avoid the existing gamepad-nav wiring. Every
existing Forms control (`Button`, `ComboBox`, `ListBox`, `ScrollViewer`, `Slider`, `TextBox`, etc.)
still only does index-based Tab navigation automatically — spatial nav isn't a real, declarative
option yet.

## Remaining work

- Decide the final interface: how does a screen/container opt a scattered layout into spatial
  navigation without every existing control needing code changes? (Leading idea: a declarative
  marker — e.g. a property on the navigation root — that `HandleGamepadNavigation` itself checks,
  so every control that already calls it gets spatial nav for free when scoped under that root.)
- Keyboard Tab stays index-based always — this only affects gamepad D-pad/stick navigation.
- Explicit per-element directional override (Unity `NextUp/Down/Left/Right`-style escape hatch),
  since spatial heuristics get scattered/irregular layouts wrong sometimes.
- Finalize scoring constants/API surface based on the spike's manual-test feedback.

vchelaru:
Closes #4129

Real, integrated spatial (any-angle) focus navigation for gamepad/stick input on scattered UI layouts, per the follow-up discussion after the initial spike validated the approach.

**`FrameworkElement.GamepadNavigationMode`** (`TabOrder`/`Spatial`, inherited via the existing `ParentFrameworkElement` walk, nearest-ancestor-wins with explicit `TabOrder` opt-out) — set once on a container, every stock control (Button, ComboBox, ListBox, ScrollViewer, Slider, TextBox) gets spatial nav automatically through the same `HandleGamepadNavigation` call it already makes. No subclassing or manual gamepad wiring needed.

**`SpatialNavigationService`** — distance + angular-alignment scoring (Unity-style, not WICG's vertex decomposition) over on-screen candidate positions. Excludes the origin and any ancestor of the origin (a large focusable container wrapping the origin near an edge would otherwise win by proximity).

**Explicit per-direction override escape hatch** — `SpatialNavigationUp/Down/Left/Right`, checked before scoring for a clean single-direction D-pad press.

Keyboard Tab and FRB are both untouched.

Bugs found and fixed while testing against every control that calls `HandleGamepadNavigation`:
- Diagonal D-pad detection required both directions pushed on the exact same frame.
- Spatial mode ignored `IsUsingLeftAndRightGamepadDirectionsForNavigation` (Slider disables it so Left/Right can adjust `Value` instead of navigating).
- Ancestor-container-wins-by-proximity (found via `ScrollViewer.DoItemsHaveFocus`, which hands real focus to the child, unlike `ListBox.DoListItemsHaveFocus`'s virtual flag).

## Test plan
- Full `MonoGameGum.Tests` suite green (1997), FRB canary clean (checked after each `FrameworkElement.cs` change).
- Unit tests for: stock Button+Panel integration, no-mode regression guard, nested TabOrder opt-out, explicit override precedence/non-application on diagonal, Slider's Left/Right dual-purpose input, ListBox/ScrollViewer internal-item-focus isolation, ancestor-exclusion in scoring.
- Docs added to `tabbing-moving-focus.md`.
- Manual: scattered 12-button demo with a real gamepad, still in progress.

### DIFF
--- a/GumCommon/GumCommon.csproj
+++ b/GumCommon/GumCommon.csproj
@@ -288,6 +288,7 @@
 		<Compile Include="..\MonoGameGum\Forms\Controls\FrameworkElementExt.cs" Link="Forms\Controls\FrameworkElementExt.cs" />
 		<Compile Include="..\MonoGameGum\Forms\FrameworkElementTemplate.cs" Link="Forms\FrameworkElementTemplate.cs" />
 		<Compile Include="..\MonoGameGum\Forms\FrameworkElementTreeExtensions.cs" Link="Forms\FrameworkElementTreeExtensions.cs" />
+		<Compile Include="..\MonoGameGum\Forms\SpatialNavigationService.cs" Link="Forms\SpatialNavigationService.cs" />
 		<Compile Include="..\MonoGameGum\Forms\Controls\Button.cs" Link="Forms\Controls\Button.cs" />
 		<Compile Include="..\MonoGameGum\Forms\Controls\CheckBox.cs" Link="Forms\Controls\CheckBox.cs" />
 		<Compile Include="..\MonoGameGum\Forms\Controls\ColorPicker.cs" Link="Forms\Controls\ColorPicker.cs" />
--- a/MonoGameGum.Tests/Forms/ControllerSpatialNavigationTests.cs
+++ b/MonoGameGum.Tests/Forms/ControllerSpatialNavigationTests.cs
@@ -1,0 +1,135 @@
+using Gum.Forms.Controls;
+using Gum.GueDeriving;
+using Gum.Input;
+using Gum.Wireframe;
+
+using Shouldly;
+using Xunit;
+
+namespace MonoGameGum.Tests.Forms;
+
+/// <summary>
+/// Tests for direction-agnostic ("spatial") gamepad focus navigation — the issue #4129 spike.
+/// Exercises <see cref="FrameworkElement.HandleGamepadSpatialNavigation"/>, driven by a real
+/// <see cref="GamePad"/> data holder. Mirrors the harness approach in
+/// <see cref="RaylibGum.Tests.Forms.ControllerTabNavigationTests"/> for the index-based counterpart.
+/// </summary>
+public class ControllerSpatialNavigationTests : BaseTestClass
+{
+    private sealed class SpatialNavHarness : FrameworkElement, IInputReceiver
+    {
+        public SpatialNavHarness(InteractiveGue visual) : base(visual) { }
+
+        public void InvokeHandleGamepadSpatialNavigation(IGamePad gamepad, GraphicalUiElement navigationRoot) =>
+            HandleGamepadSpatialNavigation(gamepad, navigationRoot);
+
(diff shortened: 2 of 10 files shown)


vchelaru:
Two real bugs found during manual testing, both fixed and pinned with tests (commit 4b7b5eccd):

1. **Diagonal D-pad detection required same-frame simultaneity.** Direction was read from each button's own `ButtonRepeatRate`, which only pulses on that button's own schedule — holding Up then pressing Right shortly after (normal controller use, not exact-frame-simultaneous) could drop Up's contribution. Fixed: direction now comes from `ButtonDown` (raw held state), firing still gated on `ButtonRepeatRate`.
2. **Integration hazard, not a code bug**: a manual-test demo that also called `GumService.UseGamepadDefaults()` double-navigated, because stock `Button`s already read `GamePadsForUiControl` in their own `OnFocusUpdate` and ran the existing index-based nav off the same gamepad. Documented as a caution on `HandleGamepadSpatialNavigation`'s XML doc — a real consumer must not register the same gamepad in both places.

vchelaru:
Real integration landed (commit 5b9103ff4), per the follow-up discussion — issue #4129 retitled from spike to the committed feature.

- `FrameworkElement.GamepadNavigationMode` (`TabOrder`/`Spatial`, inherited via the existing `ParentFrameworkElement` walk, nearest-ancestor-wins with explicit opt-out support) — set once on a container, every stock control automatically gets spatial nav through the same `HandleGamepadNavigation` call it already makes. No more manual subclassing/wiring.
- Explicit per-direction override escape hatch: `SpatialNavigationUp/Down/Left/Right`, checked before scoring for a clean single-direction D-pad press.
- Keyboard Tab untouched — always index-based.
- FRB untouched — `#if !FRB` island inside the shared `HandleGamepadNavigation`, same pattern `HandleTab` already uses.
- 5 new tests proving zero-code-change stock `Button`+`Panel` integration, the no-mode regression guard, nested opt-out, and override precedence/non-application. Full suite (1992) + FRB canary both green.

Demo updated to use the real API (one `Panel.GamepadNavigationMode = Spatial`, plain stock `Button`s, no subclass) — ready for another manual pass.

Issue #2998, opened 2026-05-29
Cull off-screen renderables in Renderer.Draw (skip drawing items fully outside the active clip rect)


vchelaru:
## Problem

When a `ListBox` (or any scrolled/clipped container) holds many items, Gum draws every item every frame. The `ClipContainer`'s scissor rect clips them *visually*, but each off-screen item is still:

- walked and queued in the recursive `Renderer.Draw` pass,
- charged for any per-item custom batch (e.g. Apos.Shapes backgrounds),
- charged for visual state application and its own nested clip region.

For large lists this is significant wasted work for content that never reaches the framebuffer.

## Proposed approach

Add a **general render-walk cull** in the XNA-family renderer (`RenderingLibrary/Graphics/Renderer.cs`): when a clip/scissor rect is active, skip drawing any renderable whose bounds fall entirely outside that rect. Because this lives in the core walk, it benefits `ListBox`, `ScrollViewer`, `ComboBox` dropdowns, and any `IsClipping` container automatically.

Scope is **render-only**: skip just the draw work for off-screen items. Layout and visual state continue to run as today, so an item re-entering the view is already correct. (Skipping per-frame layout/state, and full virtualization where off-screen visuals aren't created at all, are out of scope here — track virtualization separately if desired.)

## Implementation notes / starting points

- The recursive walk and clip handling live in `RenderingLibrary/Graphics/Renderer.cs` (`Draw` / `Render` recursion, `AdjustRenderStates`, the clip-change paths). The active clip is available via `SpriteRenderer.CurrentScissorRectangle`.
- Forms scrolled content is `ScrollViewer.InnerPanel` inside `ScrollViewer.ClipContainer` (`MonoGameGum/Forms/Controls/ScrollViewer.cs`).
- This is the XNA-family path only (MonoGame/KNI/FNA). Skia/Raylib/Sokol have their own renderers and are out of scope for this issue.

## Open questions (resolve at implementation time)

- **Bleed-outside-bounds.** Children/visuals can extend past their own bounds (dropshadows, negative offsets, rotation), and a culled parent may contain children that draw outside the parent's rect. If culling decisions are based on a top-parent's bounds, those cases could be wrongly skipped. Need a safe bounds basis (e.g. cull on the renderable's own clip-relevant bounds, and be conservative about recursing into a culled subtree).
- **When to cull.** Only when a clip rect is actually active — never cull unclipped content.
[...]

PR #4152, opened 2026-07-31
Fix off-screen cull skipping scrolled multi-line TextBox text

### LINKED ISSUE #4144: Long text in a multi-line TextBox shouldn't cause caret issues and invisible text
If you type "enough" text in a multiline box, the caret ends up in the wrong place with missing/invisible text.

I don't know exactly how much "enough" is or whether "enough" depends on the size of the TextBox.

See repro video here:

https://github.com/user-attachments/assets/bb9eff27-6b4e-4e45-86a2-06d245c3e917

vchelaru:
## Summary
- The off-screen render cull (#2998) judges a renderable's own declared bounds against the active clip. A multi-line Forms TextBox scrolls its Text by moving Y while Height stays fixed to the visible box (RelativeToParent), so its declared bounds drift entirely outside the clip once scrolled far enough — culling the whole box even though the current lines should draw. Matches #4144's reported invisible-text/caret behavior.
- `GetCullTestBoundsFor` (RenderingLibrary/Camera.cs) expands a wrapped-text renderable's bounds to its actual `WrappedTextHeight` when that exceeds declared `Height`, used only for the cull decision in `HierarchicalOrderer`/`BatchKeyGroupedOrderer`. `GetScissorRectangleFor`, which drives the real GPU clip, is unchanged — so this doesn't affect what's visually cropped, only what gets culled.
- Not raylib-specific: the fix lives in shared `RenderingLibrary` code used by every backend.

## Test plan
- [x] New unit tests reproduce the bug pre-fix (compile-time proof the method didn't exist, then red/green via `HierarchicalOrdererTests`/`CameraTests`)
- [x] Regression guard confirms ordinary off-screen content (ListBox-style) is still culled
- [x] Full `MonoGameGum.Tests` suite green (2008 tests)
- [x] `RaylibGum.Tests` green (554 tests)
- [x] `OffscreenCullRenderTests` (real MonoGame draw) still green — cull still reduces draw states
- [x] RaylibGum/SkiaGum build clean
- [x] FRB canary: pass (built clean from primary checkout)
- [x] Manual: repro copy built and VS opened for the user against the raylib AMA template — pending user confirmation

Closes #4144

### DIFF
--- a/GumRuntime/GraphicalUiElement.cs
+++ b/GumRuntime/GraphicalUiElement.cs
@@ -81,7 +81,7 @@ public enum Dock
 /// wrap an underlying rendering object.
 /// GraphicalUiElements are also considered "Visuals" for Forms objects such as Button and TextBox.
 /// </summary>
-public partial class GraphicalUiElement : IRenderableIpso, IVisible, INotifyPropertyChanged
+public partial class GraphicalUiElement : IRenderableIpso, IVisible, INotifyPropertyChanged, IHasRenderableComponent
 {
     #region Enums/Internal Classes
 
--- a/MonoGameGum.Tests/RenderingLibraries/CameraTests.cs
+++ b/MonoGameGum.Tests/RenderingLibraries/CameraTests.cs
@@ -204,6 +204,110 @@ public class CameraTests : BaseTestClass
         scissor.Bottom.ShouldBe(650);
     }
 
+    // Minimal IRenderableIpso + IWrappedText double for GetCullTestBoundsFor (#4144): only
+    // Height (IRenderableIpso) and WrappedTextHeight (IText, via IWrappedText) are meaningful;
+    // every other IWrappedText/IText member is unused by the cull-bounds computation and stubbed.
+    private sealed class StubWrappedTextRenderable : IRenderableIpso, IWrappedText
+    {
+        public StubWrappedTextRenderable()
+        {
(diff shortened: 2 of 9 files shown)


vchelaru:
Manual test correctly caught that the first pass didn't work. Root cause was two additional gaps:

1. The cull walk operates on the GraphicalUiElement/TextRuntime wrapper, not the raw Text renderable — and the wrapper doesn't itself implement IWrappedText, only the inner renderable does. So `ipso is IWrappedText` silently never matched a real TextBox. Added IHasRenderableComponent so GraphicalUiElement exposes its inner renderable, and unwrap through it.
2. raylib doesn't use HierarchicalOrderer at all — it has its own separate draw walk with its own copy of the cull check, still calling the old method. Fixed that too (both its normal and render-target-bake branches).

Also caught and fixed a test-geometry bug: placing the clip container at the camera's Y=0 edge made an off-clip renderable's declared bounds collapse to a degenerate rect that spuriously read as "not outside" regardless of any fix — which is how the first PR's tests passed without actually proving anything. Re-verified red-before/green-after with corrected geometry this time, including a new raylib-specific integration test (none existed before).

Full RaylibGum.Tests (556), MonoGameGum.Tests (2009), and the MonoGame OffscreenCullRenderTests integration test all green. FRB canary: pass.

Issue #4096, opened 2026-07-29
Layer: add a per-layer screen transform + render-target awareness — needed for simultaneous UI coordinate spaces (e.g. RT-scaled game UI + window-pinned editor UI)


vchelaru:
## Problem

Common pattern: game UI drawn into a `RenderTarget2D` (then scaled/letterboxed onto the backbuffer via the game's own code — e.g. MonoGame.Extended's `ViewportAdapter`), while a separate editor/tool UI draws directly onto the backbuffer at 1:1 window pixels, on top. Both need correct mouse interaction simultaneously — a game button and an editor button both clickable in the same frame, at different scales.

Gum has no visibility into an externally-managed render target — `GraphicsDevice.SetRenderTarget(...)`, drawing Gum (or `GumBatch`) into it, then `SetRenderTarget(null)`, is entirely orchestrated by the caller. The only bridge Gum offers is `Cursor.TransformMatrix`, applied once, globally, to `Cursor.X`/`Y` before any per-element logic runs (`MonoGameGum/Input/Cursor.cs:159,566-570`). One matrix can't simultaneously undo an external RT scale for game elements and stay identity for window-pinned editor elements drawn in the same frame.

This is distinct from `ContainerRuntime.IsRenderTarget`, Gum's own native "bake this subtree to an internal texture" mechanism — that path already has its own acknowledged gap: content rendered through it isn't cursor-interactive (`docs/code/standard-visuals/spriteruntime/rendertargettexturesource.md:70-72` — "Future versions of Gum may provide a way to interact with invisible controls rendered to a RenderTarget"). Both gaps are real but separate; this issue is scoped to the externally-managed-RT case.

**Routing which content goes to which render target does not need a new mechanism.** `GumBatch.Draw(renderable)` already draws an arbitrary container subtree explicitly — a caller can already do `gumBatch.Draw(gameContainer)` while an RT is bound and `gumBatch.Draw(editorContainer)` after, with no `Layer` involved. See `Samples/MonoGameGumImmediateMode/MonoGameGumImmediateMode/Screens/RenderTargetScreen.cs` for the existing (non-interactive) version of this pattern.

[...]

PR #4162, opened 2026-07-31
Add HitTestTransformMatrix sample to RenderTarget screen

vchelaru:
## Summary
- Extends the MonoGameGumImmediateMode `RenderTarget` screen with real Forms controls (Button/Window/Menu) demonstrating `HitTestTransformMatrix` (#4096) end to end — docs previously only had a code snippet.
- Fixes the sample's `KernSmith.MonoGameGum` reference (NuGet PackageReference → ProjectReference) which threw a runtime `FileNotFoundException` when built via VS.
- Fixes a general footgun found while building this: calling `GumService.Default.Update` more than once per frame silently breaks Push/Click. Screens now expose `AdditionalUpdateRoots`; `Game1` combines Root + extra roots into one `Update` call. Documented in `mouse-and-touch-screen-cursor.md`.

## Status — draft, not done
- Known open issue: the game button intermittently stops responding depending on how `_aboveGameUiContainer` is sized (a large/`Dock.Fill`ed `ContainerRuntime` with default `HasEvents=true` can eclipse another root's controls). Still being tracked down.
- Considering a follow-up doc/skill update for the temporal Update-once-per-frame rule and the eclipsing-container gotcha.

## Test plan
- [x] `dotnet build` on the sample project — succeeds
- [x] Manual test in VS (hover/click/drag on all three controls) — pending, blocked on the open issue above


vchelaru:
FRB canary: pass. Built `FlatRedBall.Forms.DesktopGlNet6.csproj` (MenuItem.cs/Menu.cs are FRB-shared) — baseline on origin/main and this branch both 0 errors.

Issue #4176, opened 2026-07-31
Rectangle/Circle Fill*/Stroke* v3 shape properties silently no-op when loaded from a project on raylib


vchelaru:
## Problem

`RectangleRuntime`/`CircleRuntime`'s v3 shape properties (`IsFilled`, `FillRed`, `FillGreen`,
`FillBlue`, `FillAlpha`, `StrokeRed`/`Green`/`Blue`/`Alpha`, `CustomRadius*`) silently no-op when
loaded from a saved `.gumx` project on **raylib**. `TrySetPropertyOnRectangleRuntime` /
`TrySetPropertyOnCircleRuntime` in `Gum/Wireframe/CustomSetPropertyOnRenderable.cs` are gated
`#if !RAYLIB && !FRB` — every other backend (MonoGame/KNI/FNA/Skia) gets these cases; raylib falls
through to the generic reflection fallback, which — per the code's own comment at that gate —
reflects on the *contained renderable* (`Gum.Renderables.LineRectangle`/`LineCircle`), not the
wrapping runtime, and those v3 properties only exist on the runtime. The assignment is dropped with
no error.

Confirmed via a new `RaylibScreenshotService` test (#4174): a Rectangle instance with
`IsFilled = true` renders (opaque white, the property's own hard-coded default), but setting
`FillRed`/`FillGreen`/`FillBlue`/`FillAlpha` to a custom color has zero effect — the shape stays
white regardless of the requested fill color, when loaded from a project file.

This is plausibly the actual mechanism behind #4172 (raylib blend modes not matching the tool) —
a saved project's custom fill color is silently dropped on raylib, which would look exactly like
"raylib renders differently than the tool."

## Suggested improvement

Add a raylib-appropriate `TrySetPropertyOnRectangleRuntime`/`TrySetPropertyOnCircleRuntime` (or
equivalent case coverage) so the v3 shape properties apply on raylib the same way they do on every
other backend. Requires understanding why raylib was excluded when this dispatch was added (issue
#2925/#2768) — check whether `Gum.Renderables.LineRectangle`/`LineCircle` on raylib need their own
storage for these values or can reuse the same runtime-level fields the other backends' cases write.

## Source

- `Gum/Wireframe/CustomSetPropertyOnRenderable.cs` — `#if !RAYLIB && !FRB` gate (~line 3279) wrapping
  `TrySetPropertyOnRectangleRuntime` (~line 3363) and the equivalent Circle method (~line 3320s).
[...]

PR #4177, opened 2026-07-31
Add MonoGame-vs-raylib screenshot diffing (gumcli diff-screenshots)

### LINKED ISSUE #4172: Gum blend modes in Raylib should match tooling
Observe the difference between the tool preview and the in-game rendering:

Image

I had a similar issue with things rendering incorrectly in Raylib and fixed with this PR for reference:
https://github.com/profexorgeek/Airpig/pull/119

### LINKED ISSUE #4174: Add raylib screenshot backend + cross-runtime pixel-diff coverage for gumcli screenshot
## Problem

#4172 (raylib blend modes don't match tooling) slipped through because runtime blend-mode tests
(`RaylibGum.Tests/Rendering/BlendModeRenderTests.cs`, `SkiaGum.Tests/Renderables/BlendTests.cs`) each
assert hand-derived expected pixel values per-runtime. Nothing cross-checks against what the tool
actually renders (MonoGame/XNA, via `Tool/EditorTabPlugin_XNA`), so two runtimes can each pass their
own self-consistent tests while visually disagreeing.

`gumcli screenshot` already renders a `.gumx` Screen/Component to a pixel-accurate PNG via MonoGame
DesktopGL (`Tools/Gum.ProjectServices.MonoGame/MonoGameScreenshotService.cs`, behind
`Tools/Gum.ProjectServices/Screenshot/IScreenshotService.cs`). There is no raylib equivalent.

## Suggested improvement

- Add `Gum.ProjectServices.Raylib` + `RaylibScreenshotService : IScreenshotService` (same interface
  `MonoGameScreenshotService` implements).
- Add `--backend` to `gumcli screenshot` (`monogame` default, `raylib`) — anticipated already in
  `ScreenshotCommand.cs`'s doc comment.
- Add a render-all mode: enumerate every Screen/Component in a `.gumx` and render each with both
  backends to matched output paths.
- Extract `PixelComparer`/`GoldenImageAssert` (`Tests/SkiaGum.Tests/GoldenImages/`) into a shared,
  dependency-free helper; add a diff step that pairs up the two backends' renders per element and
  reports out-of-tolerance pixels (tolerance needed for AA/sampling differences — this is not a
  0-diff check).
- Wire into CI (Windows + Mesa llvmpipe, the same trick already proven for
  `RaylibGum.Tests`/`Native-AOT-Smoke`) so a real project's full screen/component set is diffed
  cross-runtime on every PR.

## Source

- `Tools/Gum.ProjectServices/Screenshot/IScreenshotService.cs` — interface to implement for raylib.
- `Tools/Gum.ProjectServices.MonoGame/MonoGameScreenshotService.cs` — reference implementation.
- `Tools/Gum.Cli/Commands/ScreenshotCommand.cs` — where `--backend` slots in.
- `Tests/SkiaGum.Tests/GoldenImages/GoldenImageAssert.cs` + `PixelComparer` — comparer to extract/reuse.
- `.github/workflows/build-and-test.yaml` — llvmpipe wiring pattern (`RaylibGum.Tests`, `Native-AOT-Smoke`).

vchelaru:
## Summary

Built cross-runtime (MonoGame vs raylib) screenshot-diffing tooling for Gum, to catch the class of
bug in #4172 (raylib rendering silently diverging from the tool) automatically instead of relying on
manual visual inspection. Part of #4174.

- `Gum.ProjectServices.Raylib` / `RaylibScreenshotService : IScreenshotService` — `gumcli screenshot
  --backend {monogame|raylib}`
- `gumcli diff-screenshots <project.gumx>` — renders every Screen/Component via both backends and
  reports mismatches, with a side-by-side HTML report (MonoGame left / raylib right, mismatches
  sorted first)
- `PixelComparer.CompareApproximate` (new `Gum.ImageDiff` project, shared with `SkiaGum.Tests`'
  golden-image tests) — proximity-tolerant comparison: a pixel only counts as a real mismatch if no
  pixel within a small search radius (default 1px, `--proximity`) matches in color, so a 1px
  antialiasing/rounding shift between renderers doesn't get reported as a divergence. Reports
  mismatch count/percentage + bounding box instead of a single first-differing pixel.
- Wired into CI (Windows, Mesa llvmpipe) alongside `RaylibGum.Tests`

**Two real bugs found and fixed while validating this against a real production project
(Airpig.Engine/Content/Gum/Gum.gumx)**, tracked as #4176:
1. `RaylibScreenshotService` never set `CanvasWidth`/`CanvasHeight` before layout — any element
   positioned via `PixelsFromMiddle`/`Percentage` (e.g. a centered dialog) rendered pinned to the
   top-left corner on raylib instead of centered.
2. `CustomSetPropertyOnRenderable`'s RectangleRuntime/CircleRuntime property dispatch was gated
   `#if !RAYLIB` entirely, so v3 shape properties (`IsFilled`/`FillRed`/`FillGreen`/`FillBlue`/
   `Stroke*`/`Blend`/`Color`) silently no-op when loaded from a saved project on raylib.

**Does NOT fix or close #4172.** After both bug fixes and the proximity-tolerant comparer, running
`diff-screenshots` against Airpig goes from 45/73 to 9/73 mismatched elements, and the remaining 9
are all tiny (<0.01% of pixels, small bounding boxes) — visually nothing like the stark blend-mode
difference in #4172's screenshot. Two likely explanations, left for follow-up: (a) `diff-screenshots`
only renders each Screen/Component's **default state** — if #4172's bug only manifests in a specific
runtime state/instance/blend combination, this sweep wouldn't surface it; (b) the actual repro
project from #4172 hasn't been obtained yet. #4172 is intentionally left open for a fresh
investigation, ideally using this new tooling against the actual reported repro.

## Test plan

- [x] `Tests/Gum.ProjectServices.Raylib.Tests` — real raylib window + render-to-texture + PNG
      export/decode; canvas-centering and Fill*/Color dispatch regressions each pinned red-then-green
- [x] `Tests/RaylibGum.Tests/Runtimes/CustomSetPropertyOnRenderableRectangleCircleColorTests.cs` —
      direct dispatch-level pin for the `Color` fix
- [x] `Tests/Gum.ImageDiff.Tests` — proximity comparer: identical/shifted/missing-content/dimension-
      mismatch cases
- [x] `Tests/Gum.Cli.Tests` — `--backend raylib`, `diff-screenshots` end-to-end (both backends in one
      process), unknown backend exits 2
- [x] Full suites green: `RaylibGum.Tests` (567), `Gum.Cli.Tests` (61), `Gum.ProjectServices.Tests`
      (434), `Gum.ImageDiff.Tests` (9)
- [x] CI: both new projects added to `AllLibraries.sln`, wired into `build-and-test.yaml`'s Mesa
      llvmpipe step

Manual test: not needed — the automated tests read back real rendered pixels from a real GPU context
and assert on them, which is a stronger check than eyeballing a screenshot.

FRB canary: verified clean (0 errors) from the primary checkout, both before and after the
`CustomSetPropertyOnRenderable.cs` dispatch change (the only FRB-shared file touched).

### DIFF
--- a/.claude/skills/gum-cross-platform-unification/SKILL.md
+++ b/.claude/skills/gum-cross-platform-unification/SKILL.md
@@ -19,6 +19,10 @@ Why a different home: these runtimes wrap Skia-specific renderables on one side 
 
 **Scope of this pair — do not over-generalize.** This Apos↔Skia pairing covers only the **Apos-specific** runtimes (`RoundedRectangleRuntime`, `ColoredCircleRuntime`, `ArcRuntime`, `LineRuntime`). It is **not** a statement that shape support is MonoGame/Skia-only or that raylib lacks shapes. The general-purpose `RectangleRuntime`/`CircleRuntime` are unified on the normal `MonoGameGum/GueDeriving/` axis (`#if RAYLIB`/`SOKOL`/`SKIA`/`XNALIKE`) and reach full filled/rounded/shadowed capability on **every** backend — raylib included (it wraps the fully-featured `LineRectangle`/`LineCircle`). See [gum-runtime-topology](../gum-runtime-topology/SKILL.md) "Shapes are NOT MonoGame/Skia-only" for the per-backend renderable table.
 
+## Don't Route Tests Around a Divergence You Find — Fix or Surface It
+
+A per-platform `#if`-gated dispatch (e.g. `CustomSetPropertyOnRenderable.cs`'s `TrySetPropertyOn*Runtime` methods) can silently drop a property for one backend: the gate skips it, reflection falls back to the low-level renderable, finds no matching member, and no-ops with no exception. Don't write a test that passes by asserting that fallback's default instead of the property actually applying — that hides the divergence. Fix the gate (see "Disagreements Are the Whole Job" below) or write a failing/pinned test that names the gap.
+
 ## Disagreements Are the Whole Job
 
 When three files diverge, every difference falls into one of two buckets:
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -397,13 +397,18 @@ jobs:
       #   - Gum.ProjectServices.SkiaGum.Tests (SkiaGum SVG export — renders into
       #                                     SKSvgCanvas, CPU-only like SkiaGum.Tests)
       #   - Gum.Analyzers.Tests            (Roslyn analyzers; pure compilation)
-      #   - Gum.Cli.Tests                  (CLI subcommand smoke tests; fonts and
-      #                                     the raylib-backend screenshot test both
-      #                                     self-gate by OS at runtime — see
-      #                                     ScreenshotCommandTests, #4174 — rather
-      #                                     than being excluded via a Trait. Runs
-      #                                     after "Install Mesa" below since that
-      #                                     raylib-backend test needs it on Windows)
+      #   - Gum.Cli.Tests                  (CLI subcommand smoke tests; fonts, the
+      #                                     raylib-backend screenshot test, and the
+      #                                     diff-screenshots test (which renders via
+      #                                     BOTH MonoGame and raylib) all self-gate
+      #                                     by OS at runtime — see
+      #                                     ScreenshotCommandTests/
+      #                                     DiffScreenshotsCommandTests, #4174 —
+      #                                     rather than being excluded via a Trait.
+      #                                     Runs after "Install Mesa" below since
+      #                                     those tests need it on Windows)
+      #   - Gum.ImageDiff.Tests            (PixelComparer; pure SkiaSharp bitmap
+      #                                     comparison, no window/GL, fully headless)
       #   - SkiaGum.Tests                  (renders into an in-memory CPU raster
       #                                     SKSurface — no GPU / window / GL. It
       #                                     was wrongly excluded as "needs a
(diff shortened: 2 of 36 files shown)


vchelaru:
Pushed the diff-screenshots mode: `ScreenshotDiffService` renders every Screen/Component through both backends and diffs them via `PixelComparer` (extracted to a new `Gum.ImageDiff` project, shared with the SkiaGum golden-image tests). `gumcli diff-screenshots <project.gumx>` wires MonoGame vs raylib.

Found and fixed a real bug along the way: `RaylibScreenshotService` was reusing a persisted hidden window across calls; MonoGame's GraphicsDevice teardown in between corrupted that window's GL context, silently breaking the next raylib render whenever both backends ran in one process (exactly what diff-screenshots does). Now creates/closes its own window per call — confirmed via 4 repeat runs of the full `Gum.Cli.Tests` suite.

Ran it end-to-end against a real production project (Airpig.Engine/Content/Gum/Gum.gumx): 73 elements, 55 flagged mismatched — consistent with #4176.

vchelaru:
Added an HTML side-by-side report (`report.html` next to the rendered PNGs, always generated): one row per element, MonoGame left / raylib right, mismatches sorted first so a big project can be scrolled through visually instead of opening each PNG.

vchelaru:
Found and fixed two more real divergences while triaging the Airpig project with the new HTML report:

1. `RaylibScreenshotService` never set `CanvasWidth`/`CanvasHeight` before layout — any element positioned via PixelsFromMiddle/Percentage (e.g. a centered dialog) rendered pinned to the top-left corner on raylib instead of centered.
2. #4176's actual fix: the RectangleRuntime/CircleRuntime property dispatcher was entirely `#if !RAYLIB`-excluded. Split it so raylib gets the same v3 shape property coverage as every other backend.

Both verified with red-then-green tests, full RaylibGum.Tests (567)/Cli/ProjectServices.Raylib suites green, FRB canary clean.

Caveat: neither fix moved Airpig's mismatch count (still 45/73). Spot-checking flagged elements (e.g. Elements/KeyValuePair) shows visually identical renders, so I suspect most remaining mismatches are single-pixel antialiasing edge noise that tolerance=2 is too strict for, not real content bugs — worth revisiting the comparison tolerance before further triage.

vchelaru:
Implemented the proximity-tolerant comparer: `PixelComparer.CompareApproximate` only counts a pixel as a real mismatch if no pixel within a small search radius (default 1px, `--proximity` to adjust) matches in color — absorbs single-pixel AA/rounding shifts between renderers without masking real differences. Reports mismatch count/percentage + bounding box instead of the old single first-hit pixel.

Re-ran against Airpig: 45/73 mismatched -> 9/73. Confirms almost all of the previous flags were shift/AA noise, not real bugs.

PR #4235, opened 2026-08-01
Stop Gum.sln from building MonoGameGum/RaylibGum

vchelaru:
`Gum.sln` was building `MonoGameGum.csproj` and `RaylibGum.csproj` even though neither is a solution member. Root cause: `GumFormsPlugin.csproj` had a `ProjectReference` to `Gum.Cli.csproj` (`ReferenceOutputAssembly=false`), added only so MSBuild would build `gumcli` before the postbuild step shells out to it. But the postbuild `Exec` already runs `dotnet build Gum.Cli.csproj` explicitly (added because solution-level build order isn't guaranteed) — the `ProjectReference` was redundant, and MSBuild builds a project's references regardless of whether they're in the `.sln`, so it silently dragged `Gum.Cli` → `Gum.ProjectServices.MonoGame`/`.Raylib` → `MonoGameGum.csproj`/`RaylibGum.csproj` into every `Gum.sln` build.

Removed the redundant `ProjectReference`.

Verified: `dotnet build Gum.sln -v:normal` no longer shows `MonoGameGum.csproj`/`RaylibGum.csproj` in the top-level build graph. `dotnet build GumFull.sln` still succeeds (0 errors).

Manual test: not needed — pure build-graph change, no runtime behavior; proven by the two builds above.


vchelaru:
Reverting: this fix caused a real regression. Removing the `ProjectReference` to `Gum.Cli.csproj` didn't just drop a redundant safety net for standalone-csproj builds — it was the only thing guaranteeing Gum.Cli finishes building *before* GumFormsPlugin's postbuild runs, even inside a normal solution build (sibling projects in a .sln have no implied order without a ProjectReference edge). Without it, the postbuild's own `dotnet build Gum.Cli.csproj` call goes from a fast up-to-date no-op to a real concurrent cold build racing other MSBuild nodes for locks on shared dependencies (GumCommon, InputLibrary, etc.) — which is what broke a contributor's build with MSB3073 (exit code 1).

Branch reverted to parity with main. Filing a follow-up issue for the original problem (Gum.sln unnecessarily building MonoGameGum/RaylibGum) since it needs a different approach that doesn't touch the build-ordering guarantee.

PR #4260, opened 2026-08-02
Fix BitmapFont's space-character fallback throwing IndexOutOfRangeException

### LINKED ISSUE #4168: BitmapFont.GetCharacterInfo's space-character fallback can throw IndexOutOfRangeException, silently swallowed per-character by MeasureString
## Problem

Reported by the user while manually testing PR #4167: some sample/theme-generated projects run **incredibly slow**. Suspicion: they reference invalid fonts, which triggers a large number of exceptions.

## Likely mechanism (found via code review, not yet confirmed by running the actual sample)

`BitmapFont.MeasureString` measures every character of every line by calling `GetCharacterInfo(character)` wrapped in a `try`/`catch` that silently swallows failures:

- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:1173-1181` — per-character `try { characterInfo = GetCharacterInfo(character); } catch { /* swallowed */ }` inside `MeasureString`.
- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:364-378` — `GetCharacterInfo(int asciiNumber)` falls back to `mCharacterInfo[' ']` (index 32) for any out-of-range codepoint. If the font's character table doesn't extend to index 32 (e.g. an icon/symbol font whose codepoints are all below 32, or a font that failed to generate correctly), this indexer throws `IndexOutOfRangeException`.
- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:446-447` — the character array is sized from `parsedData.Chars.LastOrDefault()?.Id + 1`, so a font whose glyph table is small/mismatched for the text being rendered produces a small `mCharacterInfo` array.

`MeasureString` runs during every layout pass, so a Text instance whose Font doesn't actually match the characters it's rendering (or a "invalid"/misgenerated font) throws one exception **per character, per measurement** — not just once. This is caught and doesn't crash, but .NET exceptions are expensive, and first-chance exception overhead under a debugger (which the standard manual-test flow for this repo runs samples under) can make this dramatically worse — plausibly explaining "incredibly slow" without any crash.

## Suggested improvement

- Make the out-of-range/missing-character case in `GetCharacterInfo`/`MeasureString` a normal, exception-free control path (bounds-check before indexing `mCharacterInfo[' ']`, return `null` instead of throwing) instead of relying on a swallowed exception per character.
- Confirm which theme-generated sample(s) and font(s) actually trigger this — attach a debugger with "break on all exceptions" (or log `System.AppDomain.CurrentDomain.FirstChanceException`) while running one of the slow samples to verify this is in fact the hot path, and identify which font reference is invalid/mismatched so the generation step that produced it can also be fixed.

## Source

- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:364-378` (`GetCharacterInfo`)
- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:1166-1181` (`MeasureString`'s per-character try/catch)
- `RenderingLibrary/Graphics/Fonts/BitmapFont.cs:439-447` (`SetFontPattern`, character array sizing)

vchelaru:
## Problem
`BitmapFont.GetCharacterInfo` (and 5 sibling methods: `AssignCharacterTextureCoordinates`, `DistanceFromTopOfLine`, `GetCharacterHeight`, `GetCharacterScaleX`, `GetCharacterSpacing`) fell back to `mCharacterInfo[' ']` for any out-of-range codepoint lookup. That index (32) can itself be out of bounds for a font whose character table has fewer than 33 entries, throwing `IndexOutOfRangeException`. `MeasureString` silently swallowed this per character via a `try`/`catch` -- expensive under a debugger, and the exact mechanism #4168 (previously titled around the Bubblegum-theme slowdown) hypothesized.

## Fix
- Cache the resolved space glyph in a new `_defaultCharacterInfo` field, set once `SetFontPattern` finishes populating the array, instead of re-indexing `mCharacterInfo[' ']` at every fallback call site.
- Removed `MeasureString`'s now-dead `try`/`catch` (`GetCharacterInfo` can no longer throw).
- Removed a redundant re-fill of `mCharacterInfo[' ']` in `SetFontPattern` that recomputed an already-identical value.

## A regression along the way
My first pass cached `_defaultCharacterInfo` too early -- before `SetFontPattern`'s per-character loop overwrites index 32 with the font's real, final space glyph (it starts from a shared placeholder object that the tab/newline special-casing also mutates in place). Two existing tests (`WrappedText_ShouldRemoveZeroWidthSpace_WhenBreakingAtIt`, `WrappedText_ShouldHandleMultipleZeroWidthSpaces_InSameWord`) caught this by going red. Fixed by moving the assignment to after that loop, and added a dedicated regression test (`GetCharacterInfo_ShouldFallBackToRealSpaceGlyphMetrics_NotPlaceholderMutatedByTabOrNewline`) pinning it.

## Testing
- `dotnet test MonoGameGum.Tests/MonoGameGum.Tests.csproj` -- 2088 passed.
- New tests: reproduces the `IndexOutOfRangeException` for a small font (red before the fix), and pins the real-vs-placeholder space glyph regression (red without the post-loop reassignment).
- `RaylibGum.Tests` / `SkiaGum.Tests` build clean (0 errors) -- `BitmapFont.cs` is shared across these runtimes.

Manual test: not needed -- the observable behavior (fallback glyph metrics, wrapped line content) is pinned exactly by the unit tests above; a manual check would only approximate the same data by eye.

FRB canary: needed -- `BitmapFont.cs` is FRB1-shared via `GumCoreShared.projitems`. No platform `#if` or new BCL APIs touched, so risk is low, but will run once the primary checkout is free.

Closes #4168

### DIFF
--- a/MonoGameGum.Tests/RenderingLibraries/BitmapFontTests.cs
+++ b/MonoGameGum.Tests/RenderingLibraries/BitmapFontTests.cs
@@ -161,6 +161,45 @@ char id=37   x=161   y=0     width=22    height=20    xoffset=1     yoffset=6   
         font.MeasureString("     ").ShouldBe(40 + 5 + 1);
     }
 
+    const string smallCharSetFontData =
+@"info face=""Arial"" size=-18 bold=0 italic=0 charset="""" unicode=1 stretchH=100 smooth=1 aa=1 padding=0,0,0,0 spacing=1,1 outline=0
+common lineHeight=21 base=17 scaleW=256 scaleH=256 pages=1 packed=0 alphaChnl=0 redChnl=4 greenChnl=4 blueChnl=4
+page id=0 file=""Font18Arial_0.png""
+chars count=1
+char id=5   x=0   y=0   width=3     height=1     xoffset=-1    yoffset=20    xadvance=5     page=0  chnl=15
+";
+
+    [Fact]
+    public void GetCharacterInfo_ShouldNotThrow_WhenFontIsSmallerThanSpaceCharacterIndex()
+    {
+        BitmapFont font = new BitmapFont((Texture2D)null!, smallCharSetFontData);
+
+        font.SetFontPattern(256, 256);
+
+        // The font's character array only extends to index 5 (its last defined char id),
+        // short of the space character's index (32). An out-of-range lookup falls back
+        // to the space character, which used to re-index the same too-short array and throw.
+        Should.NotThrow(() => font.GetCharacterInfo('A'));
+    }
+
+    [Fact]
+    public void GetCharacterInfo_ShouldFallBackToRealSpaceGlyphMetrics_NotPlaceholderMutatedByTabOrNewline()
+    {
+        BitmapFont font = new BitmapFont((Texture2D)null!, basicBMFontFileData);
+
+        font.SetFontPattern(256, 256);
+
+        // Codepoint 200 is out of range (the font only defines up to id 37) and falls back to
+        // the space character. SetFontPattern initially fills every array slot with a shared
+        // placeholder object, then mutates that same object in place for the tab/newline special
+        // cases, before finally overwriting index 32 with the font's real space glyph (id 32
+        // above). The fallback must reflect that real, final glyph -- not the placeholder, which
+        // by then has its texture coordinates zeroed out by the newline special-case.
+        BitmapCharacterInfo fallback = font.GetCharacterInfo((char)200);
+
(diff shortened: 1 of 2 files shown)


vchelaru:
FRB canary: pass. Built `GumCore/GumCoreXnaPc/GumCore.DesktopGlNet6/GumCore.DesktopGlNet6.csproj` (net6.0 + net8.0) from the primary checkout against both `main` (baseline) and this branch — 0 errors on both.

Issue #4282, opened 2026-08-03
Evaluate ReadyToRun publishing to reduce first-call JIT warmup cost


vchelaru:
## Problem

Several operations in the Gum tool are dramatically slower on their first invocation than in steady state - e.g. during the #4280/#4281 selection-performance investigation, the first `PropertyGridManager.RefreshDataGrid` call measured 700-1000ms+ vs ~30-90ms on later calls, and the first screen/element switch measured 1.2-1.8s vs 200-400ms on subsequent switches. Part of this is plausibly JIT warmup - methods get compiled the first time they run, not ahead of time.

## Suggested improvement

Publish the Gum tool with ReadyToRun (`PublishReadyToRun=true`) so native code is precompiled ahead of time and shipped alongside the IL, reducing first-call JIT cost. This is a build/publish-time change only - no application code changes. WPF is compatible with ReadyToRun (unlike NativeAOT, which WPF's reflection-heavy XAML/BAML pipeline does not support).

**Test in isolation** from the background-warm-up issue (see companion issue) so each lever's actual contribution to cold-start time can be measured separately - don't ship both at once.

## Source

- `Gum/Gum.csproj` - publish configuration; add `PublishReadyToRun` (and `RuntimeIdentifier`, required for R2R) to the appropriate publish profile.
- Background: see the private investigation notes from #4280/#4281 for the original warmup measurements.

vchelaru:
PR #4285: gates `PublishReadyToRun=true` in `Gum/Gum.csproj` to the win-x64 Release publish, plus temporary timing instrumentation (logs Main entry / first paint / first `RefreshDataGrid` call to a file) so the before/after numbers can actually be compared. R2R confirmed to take effect in the publish output. Awaiting a manual timing run before merge; the instrumentation will be removed or kept per those results.

vchelaru:
Measured before/after (2 runs each, same commit, only `PublishReadyToRun` toggled at publish time — win-x64, self-contained false, single-file):

| | No R2R | R2R |
|---|---|---|
| Main → first `RefreshDataGrid` | 3064ms, 3129ms (avg 3097) | 2503ms, 2773ms (avg 2638) |
| Main → first paint | 3769ms, 3774ms (avg 3772) | 3104ms, 3601ms (avg 3352) |

~15-18% faster to first grid refresh, ~11% faster to first paint, consistent across both runs. Real, repeatable win for a config-only, zero-code-risk change. Keeping it — PR #4285.

PR #4284, opened 2026-08-03
Warm up hot Variables-tab WPF displayer controls at startup (#4283)

### LINKED ISSUE #4283: Evaluate background warm-up of Variables-tab WPF control types to hide first-use cost
## Problem

Same symptom as the companion ReadyToRun issue (#4282): first-use operations in the tool are far slower than steady-state - e.g. the first `PropertyGridManager.RefreshDataGrid` call measured 700-1000ms+ vs ~30-90ms steady state. Unlike the JIT-warmup hypothesis behind #4282, part of this cost is plausibly WPF-specific: the first time a given control type (`TextBoxDisplay`, `SliderDisplay`, `ComboBoxDisplay`, etc.) is instantiated, WPF resolves its style/template and parses BAML resources, which is a one-time cost per control type per app run.

## Suggested improvement

After the main window is shown (not before - must not delay app appearance), proactively instantiate one instance of each hot WPF control type used in the Variables tab on a background thread or low-priority `Dispatcher.BeginInvoke` (`DispatcherPriority.Background`/`ApplicationIdle`), so that cost lands during idle time instead of on the user's first real click into the Variables grid.

**Test in isolation** from the ReadyToRun issue (#4282) so each lever's actual contribution to cold-start/first-click time can be measured separately - don't ship both at once.

## Source

- `Tool/EditorTabPlugin_XNA/MainEditorTabPlugin.cs:830` (`HandleXnaInitialized`) - where tool startup/window-ready sequencing currently lives; the warm-up pass should hook in after the main window is visible, not block it.
- `WpfDataUi/DataUiGrid.cs` / `WpfDataUi/Controls/SingleDataUiContainer` - where the actual per-displayer-type WPF controls get created; identify the control types to warm.
- Background: see the private investigation notes from #4280/#4281 for the original warmup measurements.

vchelaru:
## Summary
Evaluates the fix proposed in #4283: proactively construct one instance of each hot Variables-tab WPF displayer control type on a background-priority dispatcher callback after the main window is shown, so the one-time style/template resolution + BAML parse cost for each type lands during startup idle time instead of on the user's first click into the Variables grid.

Hot set (confirmed via `PreferredDisplayer` assignments in `StandardElementsManager.GumTool.cs`/`PropertyGridManager.cs` and `SingleDataUiContainer`'s type-fallback table): `TextBoxDisplay`, `SliderDisplay`, `ComboBoxDisplay`, `CheckBoxDisplay`, `NullableBoolDisplay`, `AngleSelectorDisplay`, `FileSelectionDisplay`, `MultiLineTextBoxDisplay`, `StringListTextBoxDisplay`, `ListBoxDisplay`, `ToggleButtonOptionDisplay`, `ColorDisplay`, `VariableRemoveButton`.

Excluded as not confirmed hot for the Variables tab specifically: `MultiFileDisplay` (Project Properties dialog only), `PlusMinusTextBox`/`EditableComboBoxDisplay` (no call sites found anywhere in the tool).

Each warm-up construction is queued as its own `Dispatcher.BeginInvoke(DispatcherPriority.Background, ...)` call (chained one-at-a-time) rather than one big batch, so real UI work (paint, input) always gets priority over the remaining warm-up work.

Also adds **temporary** timing instrumentation (`Gum.Diagnostics.StartupPerfLog`, per the `self-serve-logging` skill) writing to `%TEMP%\gum_startup_timing.log`: time to `MainWindow.Visibility = Visible`, time to `MainWindow.ContentRendered` (first paint), warm-up schedule/completion, and the first `PropertyGridManager.RefreshDataGrid` call with how many of the 13 types were already warmed at that point. Remove after evaluation.

Companion to #4282 (ReadyToRun publish) - intentionally separate, does not touch publish/csproj settings.

Closes #4283

## Test plan
- [x] `dotnet build GumFull.sln` - succeeds, no new warnings
- [ ] Manual: run built `Gum.exe`, use the Variables tab, confirm no regression; read `%TEMP%\gum_startup_timing.log` for before/after timing

### DIFF
--- a/Gum/Diagnostics/StartupPerfLog.cs
+++ b/Gum/Diagnostics/StartupPerfLog.cs
@@ -1,0 +1,42 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace Gum.Diagnostics;
+
+/// <summary>
+/// TEMPORARY instrumentation for issue #4283 (evaluating a WPF displayer control warm-up pass).
+/// Writes wall-clock-since-start timestamps to a fixed log file so before/after startup timing
+/// can be compared without asking anyone to relay console output. Remove once #4283 is resolved.
+/// </summary>
+internal static class StartupPerfLog
+{
+    private static readonly string LogFilePath = Path.Combine(Path.GetTempPath(), "gum_startup_timing.log");
+    private static readonly Stopwatch Stopwatch = new();
+
+    /// <summary>
+    /// Starts the stopwatch and truncates the log file. Must be called once, as early as possible
+    /// in <c>Main</c>, before anything else runs.
+    /// </summary>
+    public static void Start()
+    {
+        Stopwatch.Restart();
+        File.WriteAllText(LogFilePath, string.Empty);
+    }
+
+    /// <summary>
+    /// Appends a timestamped line (milliseconds since <see cref="Start"/>) to the log file.
+    /// Never throws - a diagnostic write must not be able to affect real app behavior.
+    /// </summary>
+    public static void Log(string message)
+    {
+        try
+        {
+            File.AppendAllText(LogFilePath, $"{Stopwatch.ElapsedMilliseconds,6}ms | {message}{Environment.NewLine}");
+        }
+        catch
+        {
+            // Best-effort diagnostic logging only.
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


vchelaru:
Closing without merging — measured no benefit, see issue #4283 comment for the numbers and why.

Issue #4321, opened 2026-08-04
GameUiSamples.csproj content glob has a leading space, so PreserveNewest never applies


vchelaru:
## Problem

`Samples/GameUiSamples/GameUiSamples.csproj:35` has a stray leading space inside the item's `Update` glob:

```xml
<None Update=" Content\**\*.*">
  <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
</None>
```

The glob is `" Content\**\*.*"`, not `"Content\**\*.*"`, so it matches nothing and the `PreserveNewest` metadata never applies. The sample happens to still run (content reaches the output directory by another route), which is exactly why this has gone unnoticed — it fails silently rather than breaking the build.

Found while copying this sample into a scratch project to exercise Native AOT for #4318; the copy did not get its `Content/GumProject` folder until the space was removed.

Every sibling sample writes it correctly, e.g. `Samples/MVVM/MonoGameAndGum.csproj:33` uses `<None Update="Content\GumProject\**\*.*">`.

## Suggested improvement

Drop the space. Worth confirming afterwards which mechanism is currently copying the content, so removing the space does not start copying `Content.mgcb` / `Content/obj` into the output as a side effect — if it does, narrow the glob to `Content\GumProject\**\*.*` to match the MVVM sample.

## Source

- `Samples/GameUiSamples/GameUiSamples.csproj:35`

vchelaru:
PR #4324.

Correction to the premise: MSBuild trims whitespace from item specs, so ` Content\**\*.*` **did** match and `PreserveNewest` **was** applying. Verified with `dotnet msbuild -getItem:None` (the `.gumx` item carries `CopyToOutputDirectory: PreserveNewest`) and a clean build that copies all 237 `GumProject` files. Whatever blocked the scratch copy in #4318 had another cause.

Still fixed the space, and narrowed the glob to `Content\GumProject\**\*.*` to match Samples/MVVM — `Content\**` was also sweeping in `Content.mgcb` and the MGCB-generated `Content/bin`/`Content/obj` dirs. Output is otherwise unchanged. Fully done.

Issue #4320, opened 2026-08-04
Gum-originated trim warnings still surface in user AOT publishes (residue of #4116)


vchelaru:
## Problem

A `PublishAot` publish of a Gum game still emits ~15 unsuppressed trim-analysis warnings that originate inside Gum, not in user code. They are noise in the sense that the user cannot act on them, but they also bury the warnings the user *can* act on (the `SetBinding(string)` IL2026s, which correctly point at the lambda overload).

Measured on a Native AOT publish of a converted `Samples/MVVM` while working #4318, after that issue's fixes landed:

| Site | Code |
|---|---|
| `GumRuntime/GraphicalUiElement.Binding.cs:555` — `ApplyVmValueToUi` | IL2075 |
| `GumRuntime/GraphicalUiElement.cs:6326` — `SetPropertyThroughReflection` | IL2075 |
| `GumRuntime/GraphicalUiElementPropertyReadExtensions.cs:255` — `ResolveReflectedProperty` | IL2070 |
| `Gum/Mvvm/ViewModel.cs:116` — `ViewModel()` ctor | IL2072 |
| `MonoGameGum/Forms/BehaviorFormsPropertyApplier.cs:54` — `Apply` | IL2075 |
| `MonoGameGum/Forms/Controls/ComboBox.cs:531` — `UpdateToObject` | IL2075 |
| `MonoGameGum/Forms/Controls/FrameworkElement.cs:2045` — `ReflectionRegistration` ctor | IL2075 |
| `MonoGameGum/Forms/Controls/FrameworkElement.cs:664` — `GetGraphicalUiElementForFrameworkElement` | IL2075 |
| `MonoGameGum/Forms/Controls/ItemsControl.cs:297` — `CreateNewVisual` | IL2075 ×2 |
| `MonoGameGum/Forms/Controls/ListBox.cs:468` — `DisplayMemberPath.set` | IL2075 |
| `MonoGameGum/Forms/Controls/ListBox.cs:631` — `CallUpdateToObject` | IL2075 |
| `MonoGameGum/Forms/Controls/ListBox.cs:671` — `CreateNewListBoxItem` | IL2075 |
| `MonoGameGum/Forms/Data/NpcBindingExpression.cs:32/37` — ctor | IL2075, IL2072 |

[...]

Issue #4319, opened 2026-08-04
XML-format .fnt files still use XmlSerializer, breaking font loading under Native AOT


vchelaru:
## Problem

`ParsedFontFile` accepts three BMFont formats and routes the XML one through `XmlSerializer`:

```csharp
// RenderingLibrary/Graphics/Fonts/ParsedFontFile.cs:66
private void ParseXmlText(string contents)
{
    XmlSerializer serializer = FileManager.GetXmlSerializer(typeof(XMLFont));
    ...
}
```

`XmlSerializer` is not Native AOT compatible — the same root cause as #4170/#4171 and ADR 0013. A game that converted its project to JSON (`.gumj`) and therefore believes it is AOT-clean will still crash at font load time if any `.fnt` file is in XML format rather than the text format the Gum tool emits.

Confirmed present via trim analysis while working #4318:

```
RenderingLibrary\Graphics\Fonts\ParsedFontFile.cs(70): Trim analysis warning IL2026:
  RenderingLibrary.Graphics.ParsedFontFile.ParseXmlText(String): Using member
  'System.Xml.Serialization.XmlSerializer.Deserialize(TextReader)' which has
  'RequiresUnreferencedCodeAttribute'
```

Not reproduced at runtime — the Gum tool generates text-format `.fnt`, so both AOT labs used for #4318 took `ParsePlainText` and passed. The exposure is a user who points Gum at a `.fnt` produced by BMFont's XML export.

## Suggested improvement

`XMLFont` is a small flat DTO. Replacing `XmlSerializer` with a hand-rolled `XmlReader`/`XDocument` parse would be AOT-safe with no format change and no user-visible migration, unlike the project-file family where hand-rolling was rejected (ADR 0013) because those Save classes evolve constantly. A font file's schema is fixed by the AngelCode spec.

Failing that, at minimum detect the XML format under AOT and throw a message naming the fix ("convert this .fnt to text format"), rather than surfacing the raw `NotSupportedException: MakeGenericMethod() is not compatible with AOT compilation`.

## Source

- `RenderingLibrary/Graphics/Fonts/ParsedFontFile.cs:37` — format detection
- `RenderingLibrary/Graphics/Fonts/ParsedFontFile.cs:66` — the `XmlSerializer` path
- `Direction/decisions/0013-adopt-json-project-format-for-native-aot.md` — why XML and AOT are incompatible

vchelaru:
[...]

Issue #4116, opened 2026-07-30
Audit remaining unverified reflection sites for PublishTrimmed safety


vchelaru:
## Problem

While verifying the fix for #4104 and confirming #4115, the .NET trim analyzer (`dotnet publish -p:PublishTrimmed=true`) flagged several more unannotated reflection call sites in the same family as #4115's confirmed bug (`Type.GetProperty`/`GetField`/`GetEvent`/`GetConstructor` with no `[DynamicallyAccessedMembers]`, `Delegate.CreateDelegate` with no `[RequiresUnreferencedCode]`). These are **flagged by the analyzer but not individually verified** to actually misbehave at runtime the way #4115 was — this issue is to audit them.

Sites flagged (file:line, method):
- `GumRuntime/GraphicalUiElement.Binding.cs:489` — `ApplyVmValueToUi`, `Type.GetProperty`
- `GumRuntime/GraphicalUiElement.Binding.cs:531` — `BindEvent`, `Delegate.CreateDelegate`
- `GumRuntime/GraphicalUiElement.Binding.cs:240` — `UnsubscribeEventsOnOldViewModel`, `Type.GetEvent`
- `GumRuntime/GraphicalUiElement.Binding.cs:421` — `UpdateToVmProperty`, `Type.GetEvent`
- `GumRuntime/GraphicalUiElement.Binding.cs:442` — `UpdateToVmProperty`, `Type.GetProperty`
- `GumRuntime/GraphicalUiElement.Binding.cs:448` — `UpdateToVmProperty`, `Type.GetField`
- `GumRuntime/GraphicalUiElement.cs:6326` — `SetPropertyThroughReflection`, `Type.GetProperty`
- `GumCommon/Mvvm/PropertyPathObserver.cs:278` — `GetIndexedValue`, `Type.GetProperties`
- `GumCommon/Mvvm/PropertyPathObserver.cs:145` — `WalkSegments`, `Type.GetProperty`
- `MonoGameGum/Forms/Controls/FrameworkElement.cs:640` — `GetGraphicalUiElementForFrameworkElement`, `Type.GetConstructor` (the `[Obsolete]` `DefaultFormsComponents` legacy path)
- `Gum/DataTypes/VariableListSaveExtensionMethods.cs:11` — `GetTypeConverter`, `TypeDescriptor.GetConverter` (different mechanism — `TypeConverter` resolution, not member reflection)

[...]

PR #4322, opened 2026-08-04
Make registered runtime type instantiation trim-safe

### LINKED ISSUE #4318: Trimmed/NativeAOT builds crash instantiating runtime types: reflective (bool,bool) ctor lookup is not trim-safe
## Problem

A user publishing a KNI/MonoGame game with `PublishTrimmed` or `PublishAot` crashes on the first `ToGraphicalUiElement` call:

```
System.MissingMethodException: Cannot dynamically create an instance of type 'ListBoxRuntime'.
Reason: No parameterless constructor defined.
   at System.Activator.CreateInstance(Type type)
   at GumRuntime.ElementSaveExtensions.TryCreateStrongTypeForElement(ElementSave, Boolean, String)
   at GumRuntime.ElementSaveExtensions.CreateGueForElement(...)
   at GumRuntime.InstanceSaveExtensionMethods.ToGraphicalUiElement(...)
   at Gum.Wireframe.GraphicalUiElement.CreateChildrenRecursively(...)
```

The class is not trimmed — its **constructor** is. `ElementSaveExtensions.TryCreateStrongTypeForElement` (`GumRuntime/ElementSaveExtensions.GumRuntime.cs:160`) resolves the ctor reflectively:

```csharp
var constructorWithArgs = type.GetConstructor(new Type[] { typeof(bool), typeof(bool) });
if (constructorWithArgs != null) { ...Invoke(fullInstantiation, true)... }
else { toReturn = (GraphicalUiElement)Activator.CreateInstance(type); }
```

Generated runtime classes declare exactly one ctor — `XRuntime(bool fullInstantiation = true, bool tryCreateFormsObject = true)`. Optional parameters do not make it parameterless as far as `Activator.CreateInstance` is concerned. `typeof(XRuntime)`, emitted by codegen (`CodeGenerator.cs:1750`), roots the type but not its members, so ILLink/ILC drop the `(bool,bool)` ctor, `GetConstructor` returns null, and the fallback throws.

This is why only *some* types fail: a runtime class the game constructs directly (`new MainMenuUIRuntime()`) keeps its ctor; one only ever created through the `ElementSave` path does not.

`FormsUtilities.cs:651+` registers the built-in `DefaultFromFile*Runtime` types the same way, so this is not exclusive to generated user code.

## Suggested improvement

1. Annotate the `Type` parameter of `ElementSaveExtensions.RegisterGueInstantiationType` with `[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]`. Callers pass a literal `typeof(X)`, which the trimmer can follow, so the ctors are preserved in user assemblies and in MonoGameGum itself. The retrieval side needs an `UnconditionalSuppressMessage` — the pattern already used throughout `GraphicalUiElement.Binding.cs` and `BinderHelpers.cs`.
2. Replace the silent `GetConstructor == null` fallback with an exception that names the type and the likely cause, instead of letting `Activator` throw a message with no path to the answer.
3. Extend `Tests/NativeAotSmoke.Harness` so it instantiates codegen'd components and Forms controls from a real project, not just `GumService.Initialize` + standard elements. ADR 0013 already flags this coverage gap; the present bug is exactly what it lets through.

## Source

- `GumRuntime/ElementSaveExtensions.GumRuntime.cs:160` — reflective ctor lookup and the fallback that throws
- `Tools/Gum.ProjectServices/CodeGeneration/CodeGenerator.cs:1750` — emits `RegisterGueInstantiationType(name, typeof(X))`
- `MonoGameGum/Forms/FormsUtilities.cs:651` — same registration shape for built-in Forms visuals
- `Direction/decisions/0013-adopt-json-project-format-for-native-aot.md` — prior AOT work and the stated harness gap

vchelaru:
Fixes the reported `PublishTrimmed`/`PublishAot` crash:

```
System.MissingMethodException: No parameterless constructor defined for type 'ListBoxRuntime'.
   at GumRuntime.ElementSaveExtensions.TryCreateStrongTypeForElement(...)
```

`ElementSaveExtensions` resolved a registered type's `(bool, bool)` constructor reflectively from an unannotated `Type`. `typeof(X)` roots the type but not its members, so the trimmer removed that constructor, `GetConstructor` returned null, and the `Activator.CreateInstance` fallback threw.

## Fix

`RegisterGueInstantiationType`'s `Type` parameter now carries `[DynamicallyAccessedMembers(PublicConstructors)]`. Callers pass a literal `typeof(X)`, so the trimmer preserves the constructors at every registration site — including in user assemblies, which Gum's own `ILLink.Descriptors.xml` cannot reach. The `Dictionary<string, Type>` round-trip loses the annotation, so the read side carries a documented suppression.

The attributes are gated on `NET5_0_OR_GREATER`: this file is in `GumCoreShared.projitems`, and `GumCoreDesktopGL`/`GumCoreXnaPc` target .NET Framework 4.7.x, where these attributes do not exist. It is the only shared file that uses them, so there was no existing precedent to rely on.

Also in this PR:

- The null-constructor fallback throws a message naming the element, the type, and trimming, instead of letting `Activator` throw one with no path to the answer.
- `ItemsControl.CreateNewVisual` gained the same guard — a null constructor there produced a `NullReferenceException`.
- `RegisterRuntimeTypesThroughReflection`'s already-guarded assembly scan is suppressed rather than left warning.

## Verification

Two throwaway AOT labs built from real samples, published with `PublishAot` and run end-to-end:

| Lab | Codegen shape | Before | After |
|---|---|---|---|
| `Samples/MVVM` | unified runtime (`RegisterGueInstantiationType`) | 42 failures | pass |
| `Samples/GameUiSamples` | Forms (`RegisterGueInstantiation` lambda) | pass | pass |

`GameUiSamples` passed before the fix because the Forms codegen path registers via a trim-safe lambda. Only the unified-runtime path — the one the reporter used — was broken.

`Tests/NativeAotSmoke.Harness` now builds a component through the registered-runtime-type path, closing the gap ADR 0013 flagged. Confirmed it is a real regression test, not a passing no-op: reverting the annotation turns it red under a Native AOT publish with the new diagnostic message.

2118 unit tests green; `KniGum`, `RaylibGum`, `SkiaGum` build clean.

## Notes

`ElementSaveExtensions.TryCreateStrongTypeForElement` was never on the #4116 audit list, which is how this reached a user.

Follow-ups filed rather than folded in: #4319 (XML `.fnt` files still use `XmlSerializer`), #4320 (remaining Gum-originated trim warnings in user publishes), #4321 (a content-glob typo in `GameUiSamples.csproj` found while building the lab).

Adds a Native AOT and Trimming doc page, and brings the `gum-cli` skill's command table up to date (`convert-to-json`, `diff-screenshots`, `pack`, `stage-forms-behaviors` were all missing).

Closes #4318


vchelaru:
**FRB canary: pass.**

Two canaries were needed, not one: `ElementSaveExtensions.GumRuntime.cs` is in `GumCoreShared.projitems`, and `ItemsControl.cs` is in `FlatRedBall.Forms.Shared.projitems`.

| Canary | Baseline (main) | Branch |
|---|---|---|
| `GumCore.DesktopGlNet6` | 0 errors | 0 errors |
| `FlatRedBall.Forms.DesktopGlNet6` | 0 errors | 0 errors |

Run from the primary checkout against the FlatRedBall sibling on `NetStandard`.

The `NET5_0_OR_GREATER` gate on the trim attributes was verified separately, because neither canary covers it (both target net6.0, where the attributes exist). `GumCoreDesktopGL.csproj` (net471) is already red on `main` for unrelated reasons, all in FlatRedBall's own `FlatRedBallDesktopGL.csproj` (`MathF`, `Path.GetRelativePath` used in net472 code), so it proves nothing either way. A minimal net472 probe confirmed the gate directly: the ungated attribute fails with CS0246, the gated form compiles clean, and the other APIs used in the new code are all net472-safe.

PR #4323, opened 2026-08-04
Fix --rebuildfonts hanging on projects that need font generation

ultimateownsz:
## Summary

Fixes the `--rebuildfonts` command-line switch hanging (never exiting) whenever a project actually needs font files generated.

## Root cause

The CLI path runs before `app.Run()` — `Program.MainAsync` only reaches `app.Run()` when `ShouldExitImmediately` is false — so the WPF dispatcher never pumps while the STA thread is blocked in `GetAwaiter().GetResult()`.

Font generation is genuinely asynchronous (`Task.WhenAll` in `HeadlessFontGenerationService.GenerateMissingFontsFor`), so the continuation resumes on a background thread. The tool's `ToolFontGenerationCallbacks` then routes output through `GuiCommands.PrintOutput` → `Dispatcher.Invoke`, which blocks forever against a dispatcher that will never run. `ShouldExitImmediately` is never observed, so the process hangs instead of exiting.

## Fix

Route `--rebuildfonts` through `IHeadlessFontGenerationService` directly, with no callbacks (`HeadlessFontGenerationService` defaults a null callbacks argument to `NoOpFontGenerationCallbacks`) — the same path the standalone `GumProjectFontGenerator` console app uses. This removes all WPF/dispatcher interaction from the command-line path.

Also drops the now-unused `IFontManager` dependency from `CommandLineManager`.

No behavior change for normal GUI use; only the `--rebuildfonts` command line is affected.

## Testing

- `CommandLineManagerTests` updated: `--rebuildfonts` is now verified to invoke `IHeadlessFontGenerationService.CreateAllMissingFontFiles` with the loaded project.
- `GumToolUnitTests`: 1218 passed, 0 failed.
- `GumFull.sln` builds with 0 errors.

### DIFF
--- a/Gum/CommandLine/CommandLineManager.cs
+++ b/Gum/CommandLine/CommandLineManager.cs
@@ -8,14 +8,14 @@ using Gum.Commands;
 using ToolsUtilities;
 using Gum.Messages;
 using Gum.Extensions;
-using Gum.Services.Fonts;
+using Gum.ProjectServices.FontGeneration;
 
 namespace Gum.CommandLine
 {
     /// <inheritdoc cref="ICommandLineManager"/>
     public class CommandLineManager : ICommandLineManager
     {
-        private readonly IFontManager _fontManager;
+        private readonly IHeadlessFontGenerationService _fontGenerationService;
         private readonly IGuiCommands _guiCommands;
         private readonly IFileCommands _fileCommands;
         private readonly IMessenger _messenger;
@@ -46,13 +46,13 @@ namespace Gum.CommandLine
         #endregion
 
         public CommandLineManager(
-            IFontManager fontManager,
+            IHeadlessFontGenerationService fontGenerationService,
             IGuiCommands guiCommands,
             IFileCommands fileCommands,
             IMessenger messenger,
             IProjectManager projectManager)
         {
-            _fontManager = fontManager;
+            _fontGenerationService = fontGenerationService;
             _guiCommands = guiCommands;
             _fileCommands = fileCommands;
             _messenger = messenger;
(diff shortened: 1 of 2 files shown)


vchelaru:
Thanks for digging into this, but I don't think the fix actually changes anything. The symptom diagnosis looks right; the change is a no-op.

`CommandLineManager` resolves `IHeadlessFontGenerationService` from DI, and `Gum/Services/Builder.cs` registers it *with* `ToolFontGenerationCallbacks`:

```csharp
services.AddSingleton<IHeadlessFontGenerationService>(provider =>
    new HeadlessFontGenerationService(
        provider.GetRequiredService<IFontFileGenerator>(),
        provider.GetRequiredService<IFontGenerationCallbacks>()));
```

The `null` -> `NoOpFontGenerationCallbacks` default only applies when you construct the service yourself, which is what `GumProjectFontGenerator` does. And `FontManager.CreateAllMissingFontFiles` was already a one-line pass-through to that same singleton with the same arguments, so the old and new call paths are identical. `OnOutput` -> `GuiCommands.PrintOutput` -> `Dispatcher.Invoke` and `ShowSpinner()` are still both on the path.

There's a second cause on top of that. By the time the CLI path runs, `new App()` has installed a `DispatcherSynchronizationContext` on the STA thread and `Main` is blocked in `MainAsync(args).GetAwaiter().GetResult()`. `await Task.WhenAll(tasks)` in `GenerateMissingFontsFor` has no `ConfigureAwait(false)`, so its continuation gets posted to a dispatcher that never pumps. That hangs even with zero callbacks.

The updated test can't catch either one, since it mocks `IHeadlessFontGenerationService` and asserts the new interface is called. It passes whether or not the app hangs.

A real fix needs both halves:

1. A font service instance with genuinely no-op callbacks. Note `BmFontExeFileGenerator` and `KernSmithFileGenerator` take callbacks too, so it isn't enough to only no-op the outer service.
2. The async work kept off the dispatcher, e.g. handle `--rebuildfonts` before `new App()` exists, or `ConfigureAwait(false)` through the generation path.

`Gum.Cli`'s `FontsCommand` is a working example of the headless path if you want a reference.

vchelaru:
We discussed on Discord and this has neither been experienced by a user nor has it been proven by TDD. Until one of those two is the case, I don't feel confident merging this. I'm leaving this open for now to see if it can be reproduced, but if it's not reproduced in a few days I'll probably close this out.

vchelaru:
We discussed on discord and unless we can get a proper repro, killing this.

Issue #4365, opened 2026-08-07
Raylib font oversampling doesn't compensate inline BBCode runs (FontScale/FontSize tags)


vchelaru:
## Problem

Raylib's font-oversampling support (#4330, parity for #4317) only re-rasterizes and compensates a `TextRuntime`'s BASE font (`Gum.Renderables.Text.Font`/`OversampleCompensationScale`, applied in `DrawPlainLine`). Inline BBCode runs (`[FontScale=...]`, `[FontSize=...]`, `[IsBold=...]`, etc., resolved per-run in `ResolveRunFont`/`DrawStyledLine`) are NOT compensated -- they keep drawing at their own independently-resolved size regardless of the current oversample ratio.

This was a deliberate scope cut for #4330 (see the doc comment on `Text.OversampleCompensationScale` and the "Limitation: BBCode Inline Runs on Raylib" section added to `docs/code/files-and-fonts/font-oversampling.md`), to avoid destabilizing the line-height/baseline math `GetLineTopOffsets`/`GetLineLayoutScale` already depend on. MonoGame/KNI/FNA do not have this limitation -- their oversampling composes correctly with inline runs (per the `TextRuntime.RegenerateOversampledFont` doc comment: "so a caller's own FontScale (including inline [FontScale] BBCode runs) keeps working while oversampling is active").

## Suggested improvement

Extend `ResolveRunFont`'s scale resolution (`Runtimes/RaylibGum/Renderables/Text.cs`) to compose `OversampleCompensationScale` into the default/`[FontScale]`-tag branches (mirroring what `EffectiveFontScale`/`DrawPlainLine` already do for plain text), while auditing `GetLineTopOffsets`/`GetLineLayoutScale` for any resulting height/baseline drift -- those specifically were left untouched to keep this cut small and low-risk.

## Source

- `Runtimes/RaylibGum/Renderables/Text.cs` -- `ResolveRunFont` (per-run font/size resolution used by `DrawStyledLine`), `GetLineTopOffsets`/`GetLineLayoutScale` (multi-line vertical stacking, the fragile area this scope cut avoided touching).

## Reach

[...]

Issue #4364, opened 2026-08-07
TextRuntime.RegenerateOversampledFont leaks the previous font on every re-rasterize during continuous zoom


vchelaru:
## Problem

`TextRuntime.RegenerateOversampledFont` (MonoGame/KNI/FNA and Raylib -- issues #4317/#4330) re-rasterizes a new font every time the hysteresis gate (1px raster delta) fires during a continuous zoom gesture, but never disposes the font it's replacing. Each call swaps `ContainedText.BitmapFont`/`Font` directly without releasing the old instance.

- MonoGame: `BitmapFont` is `IDisposable`; the outgoing instance is dropped with no `Dispose()` call.
- Raylib: worse -- `Raylib_cs.Font` wraps a live GPU texture (see the "wraps an unmanaged GPU texture, so regenerating would leak VRAM" comments already in `CustomSetPropertyOnRenderable.cs`'s base-font-cache paths). `RegenerateOversampledFont` bypasses that cache entirely and creates a fresh font directly via `InMemoryFontCreator.TryCreateFont`, so a real pinch-zoom/scroll-wheel-zoom gesture can accumulate many undisposed GPU textures in one session.

## Suggested improvement

Track the previously-assigned oversampled font per `TextRuntime` (or per `Text`) and dispose/unload it when superseded by a newly-generated one -- `BitmapFont.Dispose()` on MonoGame, `Raylib_cs.Raylib.UnloadFont(...)` on Raylib. Careful not to dispose a font that's also cached/shared elsewhere (e.g. the base FontCache-keyed entry) -- only the oversampling-specific swap targets need this.

## Source

- `MonoGameGum/GueDeriving/TextRuntime.cs` -- `RegenerateOversampledFont` (`#if XNALIKE` branch calls `ContainedText.SetOversampledDisplayFont(font)`; `#else`/Raylib branch calls the same with a `Raylib_cs.Font`). Neither branch disposes the font it replaces.
- `RenderingLibrary/Graphics/Text.cs` -- `SetOversampledDisplayFont` (MonoGame) swaps `mBitmapFont` with no disposal of the outgoing value.
- `Runtimes/RaylibGum/Renderables/Text.cs` -- `SetOversampledDisplayFont` (Raylib) swaps `Font` the same way.

## Reach

[...]

Issue #4317, opened 2026-08-03
Automatic zoom-driven font oversampling without hijacking FontScale


vchelaru:
## Problem

Manual-testing #4304 (fractional `FontSize`) surfaced that "zoom the camera and see crisp text" needs more than fractional `FontSize` — and that the existing mechanism for it, `TextRuntime.RegenerateOversampledFont` (#4302), has a design flaw: it compensates for the higher-resolution raster by writing directly into the public `FontScale` property.

That makes `FontScale` unusable for anything else while oversampling is active — a user's own `[FontScale=N]` BBCode run or explicit `text.FontScale = 2` gets silently stomped. The correct shape (discussed at length in #4304) is the same principle a `Circle`'s `Radius` follows under zoom, or how Skia renders text: world-space properties (`FontSize`, `FontScale`, `Radius`, ...) are never touched by the system. Resolution/detail increases happen at a layer the user's code never sees. `RegenerateOversampledFont` also only fires on a manual trigger (press R in the sample), not automatically as zoom changes.

## What's needed

A render-time (not property-set-time) mechanism:
- Access to the current effective screen scale (camera zoom, and/or a Layer's own transform) at draw time.
- An **internal-only** resolution multiplier per `Text` instance, applied at draw time, that composes multiplicatively with the public `FontScale` rather than overwriting it.
- Logic to decide when to re-rasterize (avoid regenerating every frame for imperceptible scale deltas).
- Should request the **exact fractional** raster target (now possible via #4304's `float FontSize`/`BmfcSave.FontSize`) rather than `RegenerateOversampledFont`'s current `(int)Math.Round(...)`, to get the full benefit of fractional sizing.

Not designed in detail yet — this issue is to capture the problem and the learnings below before a fresh implementation pass.

## Relevant findings from #4304

- **KernSmith's rasterizer genuinely uses fractional precision.** `FreeTypeRasterizer` converts the requested `Size` to 26.6 fixed-point (`(int)Math.Round(size * 64f)`) and calls `FT_Set_Char_Size` (FreeType's fractional-capable sizing API, not the integer-only pixel API) — so a `float FontSize` really does reach rasterization with ~1/64-unit precision, not a pre-rounded int.
[...]

PR #4366, opened 2026-08-07
Add Raylib font oversampling (parity with #4317) + cross-backend zoom demo screen

### LINKED ISSUE #4330: Build automatic font oversampling for RaylibGum (parity with #4317)
## Problem

#4317 adds automatic, render-time font oversampling for XNALIKE (MonoGame/KNI/FNA) — a Text on
`TextRuntime` re-rasterizes itself at the current effective camera/layer zoom via
`TextRuntime.UpdateAutomaticFontOversampling`, wired to `Text.OnPreRender`, and compensates with the
internal-only `Text.OversampleCompensationScale` (composes with the public `FontScale` rather than
overwriting it).

RaylibGum has **no oversampling mechanism at all today**, flawed or otherwise — `RegenerateOversampledFont`
and the new automatic trigger are both `#if XNALIKE`-gated in `MonoGameGum/GueDeriving/TextRuntime.cs`
and simply don't exist on the Raylib build of that file. This was scoped out of #4317 to keep that PR
reviewable; this issue tracks building the Raylib equivalent as its own follow-up.

## What's needed

- A Raylib equivalent of `RegenerateOversampledFont`/`UpdateAutomaticFontOversampling`, wired to
  `IRaylibFontCreator` (RaylibGum's in-memory font generation path — the equivalent of XNALIKE's
  `IInMemoryFontCreator`/KernSmith) instead of `CustomSetPropertyOnRenderableType.InMemoryFontCreator`.
- `Text.OversampleCompensationScale`/`Text.OnPreRender` (`RenderingLibrary/Graphics/Text.cs`) already
  exist unconditionally on the shared `Text` renderable, so Raylib's `TextRuntime` can reuse them directly
  — only the TextRuntime-side raster/trigger logic needs a Raylib-specific implementation.
- The same effective-scale lookup #4317 used (`SpriteRenderer.GetZoomAndMatrix`'s zoom resolution,
  composing main `Camera.Zoom` with a `Layer`'s own `LayerCameraSettings.Zoom`) needs a Raylib-side
  equivalent — check how RaylibGum's renderer resolves camera/layer zoom before assuming the exact same
  formula applies unchanged.

## Source

- `MonoGameGum/GueDeriving/TextRuntime.cs` — `RegenerateOversampledFont`, `UpdateAutomaticFontOversampling`,
  `GetEffectiveZoom`, `ResetAutomaticOversamplingState` (all `#if XNALIKE`-gated; this is the file to add a
  Raylib branch to, or to build a parallel Raylib-specific implementation from).
- `Runtimes/RaylibGum/Renderables/IRaylibFontCreator.cs` — Raylib's in-memory font creation interface.
- `RenderingLibrary/Graphics/Text.cs` — `OversampleCompensationScale`/`OnPreRender` (already shared, no
  changes needed here for Raylib parity).

## Reach

Any Raylib-backed game using camera zoom with dynamic (KernSmith-style) font generation — the common
"zoom the camera and see crisp text" ask #4317 solves for MonoGame/KNI/FNA has no equivalent on Raylib
until this ships.

vchelaru:
## Summary

Raylib had no font-oversampling mechanism at all (issue #4330, parity for #4317/#4309). Turns out the issue's premise was slightly off: `Text.OversampleCompensationScale`/`OnPreRender` do **not** already exist unconditionally on a shared `Text` class -- Raylib has its own completely separate `Text` renderable (`Runtimes/RaylibGum/Renderables/Text.cs`, wraps `Raylib_cs.Font`) with none of that machinery. `GetEffectiveZoom`'s `Camera`/`Layer`/`LayerCameraSettings` types *are* genuinely shared (via GumCommon), so that part reuses directly.

- Ports `OversampleCompensationScale`/`MeasurementFont`/`OnPreRender` to Raylib's `Text.cs`, scoped to plain (non-BBCode) text -- inline `[FontScale]`/`[FontSize]` BBCode runs aren't compensated yet (documented limitation, filed as #4365).
- Extends `TextRuntime.RegenerateOversampledFont`'s `#if XNALIKE` gate to `XNALIKE || RAYLIB`, with a Raylib-specific body (`Raylib_cs.Font` creation via `IRaylibFontCreator`, measured-width compensation via `MeasureTextEx`).
- Fixes a bug found while writing the real-KernSmith continuous-zoom test: `GetRequiredWidthAndHeight`'s height computation used the LIVE (oversampled, growing) font's line height instead of the pinned measurement font's -- width stayed stable but height drifted. Extracted `GetLineHeightAndDescender`/`MeasurementLineHeightInPixels` to fix.
- Reset wiring in `CustomSetPropertyOnRenderable.cs`'s Raylib branch (mirrors the XNALIKE reset).
- Adds a shared `ZoomScreen` (mirrored across MonoGameGumInCode/raylib/SilkNetGum via the `TextScreen.cs`-style `#if RAYLIB`/`#elif SKIA` single-file pattern) so camera-zoom text crispness can be verified side by side on all three backends.
- Docs: updated `docs/code/files-and-fonts/font-oversampling.md` (Raylib now supported; Skia never needed it; new BBCode limitation note).

Filed as follow-ups (not fixed here): #4364 (pre-existing font-disposal gap during continuous zoom, shared by MonoGame and Raylib) and #4365 (Raylib oversampling doesn't compensate inline BBCode runs).

## Test plan

- [x] New `Tests/RaylibGum.Tests/Runtimes/TextRuntimeFontOversamplingTests.cs` (9 tests, real `KernSmithRaylibFontCreator` -- gating, hysteresis, zoom-clamp, FontSize-reset, and a continuous-zoom width/height stability sweep) -- all pass.
- [x] Full `RaylibGum.Tests` suite (585 tests) green, no regressions.
- [x] `MonoGameGum.Tests` oversampling suite (20 tests) still green.
- [x] `GumFull.sln`, `MonoGameGumInCode.csproj`, `Samples/raylib/GumTest.csproj`, `Samples/SilkNetGum/SilkNetGumSample.csproj` all build clean.
- [x] FRB canary: pass (the only FRB-shared file touched, `CustomSetPropertyOnRenderable.cs`, only got new code inside the RAYLIB-exclusive branch, which FRB never compiles).
- [ ] Manual: verify the new Zoom screen on all three samples (see PR conversation).

Closes #4330

### DIFF
--- a/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
+++ b/Gum/Wireframe/CustomSetPropertyOnRenderable.cs
@@ -2452,12 +2452,13 @@ public partial class CustomSetPropertyOnRenderable
         }
 
 #if XNALIKE && !FRB
-        // This file also compiles (unchanged) for SilkNetGum (SKIA) and, via a namespace swap, for
-        // RaylibGum -- neither has ResetAutomaticOversamplingState (oversampling is XNALIKE-only for
-        // now; Raylib parity is issue #4317's own follow-up), so this must stay XNALIKE-gated even
-        // though it's already inside the non-FRB branch above. FRB is excluded too: an FRB build that
-        // also defines XNALIKE would otherwise see textRuntime typed as the plain GraphicalUiElement
-        // from the #if FRB branch above (no TextRuntime type available there yet), not TextRuntime.
+        // This file also compiles (unchanged) for SilkNetGum (SKIA), which has no oversampling
+        // machinery at all -- so this must stay XNALIKE-gated even though it's already inside the
+        // non-FRB branch above. RaylibGum (via the namespace-swapped #else branch below) has its own
+        // equivalent reset (issue #4330, Raylib parity for #4317). FRB is excluded too: an FRB build
+        // that also defines XNALIKE would otherwise see textRuntime typed as the plain
+        // GraphicalUiElement from the #if FRB branch above (no TextRuntime type available there yet),
+        // not TextRuntime.
         if (asText.OversampleCompensationScale != 1f)
         {
             asText.OversampleCompensationScale = 1f;
@@ -2568,6 +2569,18 @@ public partial class CustomSetPropertyOnRenderable
 
         if(textRuntime != null)
         {
+            // A font property being re-resolved means whatever automatic font oversampling (issue
+            // #4330, Raylib parity for #4317) was in effect is stale: it compensated for the OLD
+            // FontSize's raster, not the new font this method is about to assign. Reset here, the
+            // single point both the direct-property-setter and string/state paths converge on, so a
+            // Font/FontSize change never leaves a stale compensation ratio applied to a newly-resolved
+            // font. Mirrors the XNALIKE branch above.
+            if (asText.OversampleCompensationScale != 1f)
+            {
+                asText.OversampleCompensationScale = 1f;
+            }
+            textRuntime.ResetAutomaticOversamplingState();
+
             // The font family and size are authoritative on the TextRuntime — both the direct
             // property setters and the string/state path write them there. Sync them onto the
             // renderable so its font-based fallbacks (and rendering) stay correct. This subsumes
--- a/MonoGameGum.Tests/RenderingLibraries/TextTests.cs
+++ b/MonoGameGum.Tests/RenderingLibraries/TextTests.cs
@@ -204,6 +204,49 @@ char id=37   x=161   y=0     width=22    height=20    xoffset=1     yoffset=6   
             customMessage: "because the effective on-screen width is FontScale composed with OversampleCompensationScale, not just one or the other");
     }
 
+    // Issue #4330 (manual-test finding): a centered Button/RadioButton label visibly shifted
+    // down-and-right once oversampling actually engaged. UpdateIpsoForRendering used to recompute its
+    // alignment-box size inline as mPreRenderWidth/Height * EffectiveFontScale (which INCLUDES
+    // OversampleCompensationScale) -- but mPreRenderWidth/Height are already measured against the
+    // STABLE pinned MeasurementFont (see WrappedTextWidth/Height, which correctly use
+    // MeasurementFontScale instead), so re-scaling them by EffectiveFontScale double-applied the
+    // compensation. The alignment box ended up a different size than EffectiveWidth/EffectiveHeight
+    // (what it's centered against), producing a wrong (non-zero when it should be zero, or wrongly
+    // sized) alignment offset. Goes through SetOversampledDisplayFont (not a raw BitmapFont
+    // assignment) specifically because that's the only path that pins MeasurementFont -- the bug only
+    // manifests once a MeasurementFont is actually pinned.
+    [Fact]
+    public void UpdateIpsoForRendering_WhenOversamplingActive_TempSizeMatchesWrappedTextSize()
+    {
+        const int baseAdvance = 10;
+        const float oversampleRatio = 2.5f;
+        const int oversampledAdvance = (int)(baseAdvance * oversampleRatio); // 25
+
+        BitmapFont baseFont = new BitmapFont((Texture2D)null!, AbcFontData(baseAdvance, lineHeight: 20));
+        baseFont.SetFontPattern(256, 256);
+        BitmapFont oversampledFont = new BitmapFont((Texture2D)null!, AbcFontData(oversampledAdvance, lineHeight: 50));
+        oversampledFont.SetFontPattern(256, 256);
+
+        Text text = new Text();
+        text.BitmapFont = baseFont;
+        text.Width = 200;
+        text.Height = 100;
+        text.HorizontalAlignment = HorizontalAlignment.Center;
+        text.VerticalAlignment = VerticalAlignment.Center;
+        text.RawText = "AB";
+
+        // Mirrors TextRuntime.RegenerateOversampledFont: pins baseFont as MeasurementFont, then swaps
+        // in the oversampled display font and its compensation.
+        text.SetOversampledDisplayFont(oversampledFont);
+        text.OversampleCompensationScale = 1f / oversampleRatio;
+
+        var tempForRendering = text.UpdateIpsoForRenderingForTests();
+
+        tempForRendering.Width.ShouldBe(text.WrappedTextWidth, tolerance: 0.001f,
+            "because the alignment box used to center the drawn text must match this Text's own reported wrapped size, not double-apply the oversample compensation");
+        tempForRendering.Height.ShouldBe(text.WrappedTextHeight, tolerance: 0.001f);
+    }
+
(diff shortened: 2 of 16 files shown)


vchelaru:
Found and fixed the root cause of all three reports: the zoom slider wrote `Camera.Zoom` on every intermediate `ValueChanged` during the drag, but the slider's own drag math converts cursor position via `XRespectingGumZoomAndBounds()`, which divides by the CURRENT `Camera.Zoom`. Since that write fed the next frame's own conversion, it's a genuine circular dependency, not just the cosmetic "self-zooming UI" it looked like:

- **Silk.NET**: spiraled into unbounded zoom, continuing even after releasing the mouse (the runaway state corrupts the thumb's push/release hit-testing).
- **MonoGame/raylib**: jittered without ever settling far from 1x -- which is also why oversampling never visibly kicked in (the achieved zoom rarely crossed `UpdateAutomaticFontOversampling`'s 1px raster-delta threshold).

Fix: commit to `Camera.Zoom` only on `ValueChangeCompleted` (once per finished drag gesture), not on every frame. The label still updates live from the slider's own `Value`.

I also tried isolating the controls on a dedicated screen-space `Layer` so they'd be immune to the zoom they drive (your "weird design" point) -- reverted after finding raylib's and SkiaGum's `Renderer` both build ONE shared camera matrix for the whole frame and ignore `LayerCameraSettings` entirely when drawing (only the XNALIKE renderer respects it), so that isolation can't actually work on 2 of 3 backends today. Filed as #4367. The controls still render through the same zoomed camera as before -- less bad now that zoom only updates once per gesture (no more mid-drag self-scaling), but they'll still visibly resize when you release the slider. Let me know if that residual behavior is still bothersome enough to want addressed differently (e.g. a totally separate secondary window, or gating this demo until #4367 lands).

Pushed to the branch -- please retest when you get a chance.

vchelaru:
Found the real bug -- my frame-rate theory was wrong, as you suspected. Your Skia-matrix hunch was exactly right.

`Runtimes/SkiaGum/RenderingLibrary/Renderer.cs`'s top-level `Draw` applied `Camera.Zoom` via `canvas.Scale(Camera.Zoom)`/`canvas.Translate(...)` directly on the persistent `SKCanvas`, with **no `Save()`/`Restore()` bracketing it**. XNALIKE passes a fresh matrix to `SpriteBatch.Begin` every frame; raylib brackets it in an explicit `BeginMode2D`/`EndMode2D` pair. Skia's `Scale`/`Translate` mutate the canvas's persistent matrix stack in place -- with nothing ever resetting it, **every single Draw call (every frame, and every layer within a frame) multiplied onto whatever the previous call had already applied.** A single `Camera.Zoom` write was enough to trigger it; my earlier slider fix (real bug, but not this one) made no difference because the accumulation happens purely from repeated rendering, not from repeated zoom writes.

Confirmed via a red test first: two consecutive `Draw()` calls at `Camera.Zoom = 2` produced a `TotalMatrix` scale of 2, then 4 (doubling every frame -- would be 8, 16... exactly matching "grew until everything was off screen"). Fixed with `canvas.Save()`/`Restore()` bracketing the camera transform in that method, plus a regression test (pixel-based, since Save/Restore correctly means the transform is no longer observable after `Draw()` returns). Full `SkiaGum.Tests` suite (422 tests) still green.

This was a pre-existing SkiaGum bug, not something my font-oversampling changes introduced -- it just needed a `Camera.Zoom != 1` held across more than one rendered frame to surface, which nothing in this codebase's Skia samples had done before the Zoom demo screen.

Kept the 60fps cap on the SilkNetGum sample loop too (harmless, makes frame-rate-scaled behavior comparable to the other samples going forward) even though it wasn't the actual fix here.

Pushed. Ready to retest.

vchelaru:
Found it, via file-based trace logging in a live running instance (thanks for offering to let me launch it myself -- much faster than round-tripping).

`TextRuntime`'s constructor assigns the backing field directly:
```csharp
_containedText = textRenderable;
```
instead of going through the `ContainedText` **property**. That property's lazy-init getter was the *only* place `OnPreRender` ever got wired:
```csharp
Text ContainedText {
    get {
        if (_containedText == null) {   // always false -- constructor already set the field
            _containedText.OnPreRender = UpdateAutomaticFontOversampling;
        }
        ...
```
Since the constructor pre-populates the field, that branch never runs for a normally-constructed `TextRuntime` -- so the entire automatic per-frame oversampling trigger from #4317 was dead code, on **both** MonoGame and Raylib (Raylib mirrors the same constructor). Confirmed directly: `Text.PreRender()` fired every frame in the running sample, but `OnPreRender` was `null` every time, and none of `RegenerateOversampledFont`/`UpdateAutomaticFontOversampling`/the `ContainedText` getter ever ran without an explicit manual call.

This is why nothing worked no matter what you checked or dragged -- it's not a Zoom-screen bug at all, it's a pre-existing #4317 defect that predates my Raylib work. It survived because every existing test (MonoGame's original suite and my new Raylib suite) calls `RegenerateOversampledFont`/`UpdateAutomaticFontOversampling` *directly*, which lazily wires `ContainedText` as a side effect of that first manual call -- never exercising the real, automatic `OnPreRender` path a running game actually depends on.

Fix: wire `OnPreRender` right where the constructor assigns the field. New regression test on both platforms (`Constructor_WiresOnPreRenderForAutomaticOversampling_WithoutAnyExplicitOversamplingCall`) constructs a plain `TextRuntime` and asserts the wiring happened with zero oversampling-specific calls -- this is the test that would have caught it originally. Full suites green (MonoGameGum.Tests 2148, RaylibGum.Tests 586).

Pushed. This should be the real fix -- ready to retest on all three.

vchelaru:
Fixed both remaining issues:

**1. Checkbox couldn't turn oversampling back off.** `UpdateAutomaticFontOversampling` early-returned when `UseFontOversampling` was false but never reverted an already-oversampled font -- it just stopped further updates, leaving the last-oversampled raster (and its compensation) stuck in place forever. The flag is meant to be a live toggle. Fixed by re-resolving through the normal font cascade (which also resets compensation/hysteresis) the first time a previously-oversampled Text sees the flag go false. One shared fix, both platforms.

**2. Centered Button/RadioButton labels shifted down-and-right.** `Text.UpdateIpsoForRendering` (MonoGame-only, `RenderingLibrary/Graphics/Text.cs`) recomputed its alignment-box size inline using `EffectiveFontScale` (includes the oversample compensation) -- but that size is already measured against the stable pinned measurement font (same quantity `WrappedTextWidth`/`WrappedTextHeight` correctly report using `MeasurementFontScale`), so it was double-applying the compensation. That made the alignment box a different size than what it's centered against, producing a wrong offset for any centered text. Fixed by reusing `WrappedTextWidth`/`WrappedTextHeight` directly instead of re-deriving the same number with the wrong scale. TDD: confirmed red (got 8, expected 20) before applying the fix.

**#3 (global effect on all text)** is still by design -- didn't touch it, since you'd asked me to hold on that pending a decision on whether you actually want per-instance scoping.

Full suites green: MonoGameGum.Tests 2150, RaylibGum.Tests 587. Pushed -- ready to retest all three issues.

Issue #2790, opened 2026-05-15
Bring two-slot fill+stroke composition to Skia CircleRuntime


vchelaru:
@
Follow-up to #2785. As part of that issue, `FillColor` / `StrokeColor` were added to `SkiaShapeRuntime` to align the public API with the shared `CircleRuntime` (the XNA-likes already use this two-slot model since #2768). The **syntax** now matches across backends, but the **semantics** do not yet on Skia:

## The gap

Skias `Circle` renderable (and Aposs, mirror-image) is **single-slot**: one renderable instance can render *either* a filled disk *or* a stroked ring, not both at the same time — controlled by `IsFilled`. The XNA-likes (MonoGame / FNA / KNI / Raylib) sidestep this by composing two renderable instances per runtime (`_fill` + `_stroke`) via `RenderableRegistry`, which is what #2768 set up.

On Skia today, a user who sets **both** `FillColor` and `StrokeColor` on a `CircleRuntime` (or any `SkiaShapeRuntime`-derived runtime) will only see one of them render. The current Skia implementation ([added in #2785](https://github.com/vchelaru/Gum/pull/PLACEHOLDER)) routes both setters through the single contained renderables `Color` + `IsFilled` toggle, so the last non-null assignment wins. This is documented on the property XML docs, but its still a visible mismatch with the XNA backends.

## Goal

Bring the two-slot composition model to Skia so `FillColor` and `StrokeColor` can render simultaneously and match the XNA-likes behavior.

## Sketch

The most direct path mirrors the XNA `_fill` / `_stroke` pattern in `MonoGameGum/GueDeriving/CircleRuntime.cs`:

1. On Skia, a shape runtime that wants both layers holds **two** `Circle` (or other shape) renderable instances — one with `IsFilled = true` (fill slot), one with `IsFilled = false` (stroke slot).
2. The fill slot is the runtimes contained object; the stroke slot is its first child. Renderer draws parent before children, so visual order is fill under stroke under user-added children — same convention as XNA.
3. The runtimes `FillColor` setter writes to the fill slots `Color` and toggles its alpha to 0 when null. `StrokeColor` setter writes to the stroke slot.
4. `PreRender` mirrors the runtimes `Width`/`Height` onto the stroke slot (the stroke isnt a layout-aware `GraphicalUiElement` and wont auto-track its parents size) — same pattern as the XNA `PreRender`.

Open questions to settle during design:
[...]

PR #4375, opened 2026-08-07
Fix raylib/SkiaGum renderers ignoring LayerCameraSettings when drawing

### LINKED ISSUE #4367: raylib/SkiaGum Renderer ignores LayerCameraSettings when drawing (screen-space layers only work for hit-testing, not rendering)
## Problem

`LayerCameraSettings` (position/zoom/`IsInScreenSpace` overrides on a `Layer`, independent of the main `Camera`) is fully honored by the XNALIKE renderer, but raylib's and SkiaGum's own `Renderer` implementations build a SINGLE camera/view matrix from the main `Camera` for the whole frame and apply it uniformly to every layer -- `LayerCameraSettings` is never consulted during actual drawing on either backend. `Layer.cs`'s `ScreenToWorld`/`WorldToScreen` (shared, used for cursor conversion/hit-testing) DO honor it, so cursor math and rendering silently disagree: a raylib or Skia project that creates a screen-space HUD layer (the documented pattern, e.g. `Samples/FontPlayground/FontPlayground.MonoGame/Game1.cs`'s `_fractionalFontSizeText` rig) gets correct hit-testing against a layer that isn't actually screen-space when drawn -- it renders zoomed/panned like everything else on the main camera.

Also, neither raylib's nor SkiaGum's `Renderer` exposes `AddLayer()`/`RemoveLayer()` (present on the XNALIKE `Renderer`), so a project can't even create additional layers on those backends without reaching into shared/internal state.

## Suggested improvement

- Add `AddLayer()`/`AddLayer(Layer)`/`RemoveLayer(Layer)` to `Runtimes/RaylibGum/RenderingLibrary/Renderer.cs` and `Runtimes/SkiaGum/RenderingLibrary/Renderer.cs`, mirroring `RenderingLibrary/Graphics/Renderer.cs`.
- Make each backend's draw loop build a per-layer camera/view matrix from `Layer.LayerCameraSettings` (falling back to the main `Camera` when unset), instead of one matrix for the whole frame -- so `IsInScreenSpace`/position/zoom overrides actually take effect when drawing, matching what `Layer.ScreenToWorld`/`WorldToScreen` already promise for hit-testing.

## Source

- `Runtimes/RaylibGum/RenderingLibrary/Renderer.cs` -- `Draw(SystemManagers, List<Layer>)` builds one `Camera2D` from `_camera` and calls `BeginMode2D` once for all layers; `RenderLayer` has no matrix-selection logic per layer.
- `Runtimes/SkiaGum/RenderingLibrary/Renderer.cs` -- same shape (single shared matrix), and no `AddLayer`/`RemoveLayer`.
- `RenderingLibrary/Graphics/Layer.cs` -- `ScreenToWorld`/`WorldToScreen` (shared, correctly honor `LayerCameraSettings.IsInScreenSpace` for hit-testing) is the spec these renderers' draw paths should also follow.
- `RenderingLibrary/Graphics/Renderer.cs` -- the XNALIKE reference implementation that already does this correctly (per-layer matrix + `AddLayer`/`RemoveLayer`).

## Reach

Any raylib or Skia project using multiple layers for camera-independent HUD/UI content (the documented `LayerCameraSettings.IsInScreenSpace` pattern) -- found while building a screen-space controls panel for the #4330 zoom-demo sample screen, which had to be abandoned on raylib/Skia specifically because of this gap.

vchelaru:
## Problem

`LayerCameraSettings` (per-layer position/zoom/`IsInScreenSpace` overrides) was honored by `Layer.ScreenToWorld`/`WorldToScreen` (hit-testing) on raylib and Skia, but neither backend's `Renderer` consulted it when actually drawing — both built one Camera2D/canvas transform from the main `Camera` for the whole frame. A screen-space HUD layer rendered zoomed/panned with everything else instead of staying fixed, while hit-testing against it was already correct.

## Fix

- `Layer.GetEffectiveTransformationMatrix`'s camera-resolution logic is extracted into a new public `Layer.GetEffectiveCamera` — the single source of truth both hit-testing and rendering now share, so they can't drift apart.
- raylib's `Renderer` now builds a per-layer `Camera2D` instead of one shared `Camera2D` for the whole frame's `BeginMode2D`, plus `AddLayer()`/`AddLayer(Layer)`/`RemoveLayer(Layer)` (mirroring XNALIKE) so a project can create additional layers.
- SkiaGum's `Renderer` threads the drawing `Layer` through its top-level `Draw` so the canvas `Scale`/`Translate` uses that layer's effective camera, plus the same `AddLayer`/`RemoveLayer`.

## Testing

- New `Layer.GetEffectiveCamera` is a pure extraction, covered by the existing 20 `LayerTests` (all pass unchanged).
- raylib: `LayerCameraSettingsRenderTests` asserts the exact `Camera2D` (`Zoom`/`Target`) fed to `BeginMode2D` for a screen-space HUD layer, mirroring the existing `ShadowBlurCameraRestoreTests` precedent of reading the native camera state directly instead of a screenshot.
- Skia: `LayerCameraSettingsRenderTests` does a real pixel readback proving a screen-space rectangle renders at its fixed screen position despite an extreme main-camera pan/zoom.
- Full `RaylibGum.Tests` (592), `SkiaGum.Tests` (425), and `MonoGameGum.Tests` (2151) suites pass.

Manual test: not needed — covered by exact-data unit tests (real pixel readback on Skia, exact native camera-matrix assertion on raylib), not approximated by eye.

FRB canary: pass (`GumCore.DesktopGlNet6.csproj` builds clean; `Layer.cs` is FRB1-shared source via `GumCoreShared.projitems`).

Closes #4367

### DIFF
--- a/MonoGameGum/GueDeriving/CircleRuntime.cs
+++ b/MonoGameGum/GueDeriving/CircleRuntime.cs
@@ -1666,6 +1666,11 @@ public class CircleRuntime : GraphicalUiElement
         // RenderableComponent on next access. The fill slot's Color/Radius/etc. were copied
         // by Circle.Clone (ICloneable, MemberwiseClone) so the clone's fill matches source.
         toReturn.containedLineCircle = null!;
+        // The fill's MemberwiseClone shallow-copies OnPreRender too, so it still points at the
+        // SOURCE runtime's RefreshShapeState. Rebind to the clone's own instance method or the
+        // clone's stroke never gets its Width/Height mirror when added top-level to a Layer
+        // (issue #4367 follow-up).
+        toReturn.ContainedRenderable.OnPreRender = toReturn.RefreshShapeState;
         // Issue #2790: drop the inherited reference to the source's stroke slot and rebuild a
         // fresh one parented to the clone's fill so the clone is fully independent.
         toReturn.ClearStrokeRenderable();
@@ -1760,6 +1765,15 @@ public class CircleRuntime : GraphicalUiElement
             // each route to their own renderable. Without this call the runtime stays on the
             // single-slot legacy model (last-non-null-setter-wins).
             SetStrokeRenderable(new ContainedCircleType());
+
+            // SetContainedObject above (shared with raylib/Sokol) doesn't wire the
+            // OnPreRender hook SetContainedShape would -- wire it explicitly here so the
+            // two-slot stroke Width/Height mirror in SkiaShapeRuntime.RefreshShapeState still
+            // runs when this runtime is a TOP-LEVEL Layer member added via AddToManagers, not
+            // just when nested under a parent (issue #4367 follow-up; see RectangleRuntime's
+            // SKIA branch, which uses SetContainedShape directly since its branch isn't shared
+            // with raylib).
+            circle.OnPreRender = RefreshShapeState;
 
             // Defaults: white fill gated off (IsFilled = false) + white stroke, so a freshly-
             // constructed runtime renders as a stroke-only outline. Because FillColor defaults
--- a/MonoGameGum/GueDeriving/CircleRuntime.cs
+++ b/MonoGameGum/GueDeriving/CircleRuntime.cs
@@ -1786,7 +1786,11 @@ public class CircleRuntime : GraphicalUiElement
             IsFilled = false;
             StrokeColor = SKColors.White;
             StrokeWidth = 1;
-            StrokeWidthUnits = DimensionUnitType.ScreenPixel;
+            // Issue #4367 follow-up: Absolute (the implicit default MonoGame/raylib's
+            // CircleRuntime already use, since they never set this explicitly) so the stroke
+            // visually thickens as the camera zooms, matching the rest of the shape and the other
+            // two backends. Mirrors the RectangleRuntime fix.
+            StrokeWidthUnits = DimensionUnitType.Absolute;
 
             // Dropshadow is off by default; pre-seed alpha + offset/blur so toggling
             // HasDropshadow = true at runtime produces a visible shadow without further setup.
--- a/MonoGameGum/GueDeriving/RectangleRuntime.cs
+++ b/MonoGameGum/GueDeriving/RectangleRuntime.cs
@@ -1904,7 +1904,12 @@ public class RectangleRuntime : GraphicalUiElement
             IsFilled = false;
             StrokeColor = SKColors.White;
             StrokeWidth = 1;
-            StrokeWidthUnits = DimensionUnitType.ScreenPixel;
+            // Issue #4367 follow-up: Absolute (the implicit default MonoGame/raylib's
+            // RectangleRuntime already use, since they never set this explicitly) so the stroke
+            // visually thickens as the camera zooms, matching the rest of the shape and the other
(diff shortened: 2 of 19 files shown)


vchelaru:
## Manual verification (added after initial review pushback)

Added a new shared `LayerCameraSettingsScreen` (mirrored into all three "big three" samples, same `<Compile Include Link>` convention as `ZoomScreen`) that puts a crimson world-space rectangle next to a blue HUD panel added directly to a screen-space `Layer`. Zoom/pan the world content with the on-screen controls and the HUD panel should stay pixel-fixed on **all three backends** — before this fix it would have drifted/scaled with the world content on raylib and SkiaGum.

- MonoGameGumInCode: nav button "Layer Camera"
- raylib gallery: nav button "Layer Camera"
- SilkNetGum: nav button "Layer Camera"

All three sample projects build clean (0 errors).

vchelaru:
## Root cause found and fixed (your hunch was right)

You were right that it was a fill+stroke rendering bug, not related to LayerCameraSettings itself.

**Root cause:** `RectangleRuntime`'s two-slot fill+stroke composition on Skia mirrors the stroke slot's Width/Height from `SkiaShapeRuntime.PreRender()` every frame. The Skia `Renderer`'s render walk calls `.PreRender()` on whatever object is actually registered as a Layer member — for a **top-level** renderable (added via `AddToManagers`, like my HUD panel) that's the raw *fill* renderable, not the GUE wrapper, so the wrapper's `PreRender()` override (and its stroke-size mirror) never ran. The stroke stayed at its tiny construction-time default size near the origin. Nested/parented rectangles don't hit this — the render walk reaches the GUE wrapper there.

**Fix:** gave the Skia renderable an `OnPreRender` hook (mirroring the existing XNALIKE `AposShapeRuntime` pattern) wired onto the fill slot in `SetContainedShape`, so the stroke mirror runs regardless of top-level vs. nested placement. Also fixed the same latent bug in `RectangleRuntime`/`CircleRuntime`/`RoundedRectangleRuntime`'s `Clone()` (the hook delegate was shallow-copied to the wrong instance).

New TDD test `RectangleRuntimeTopLevelStrokeTests` reproduces it via pixel readback and is green after the fix. Also confirmed the earlier HUD-layer leak fix (my mistake from last round) still resolves the "persists across screens" symptom.

All four suites green: SkiaGum.Tests (426), RaylibGum.Tests (592), MonoGameGum.Tests (2151), plus the three sample projects rebuild clean.

vchelaru:
## Fixed: Silk.NET stroke not thickening with zoom

Cross-backend default mismatch: `RectangleRuntime`/`CircleRuntime`'s Skia constructors explicitly defaulted `StrokeWidthUnits` to `ScreenPixel` (holds a constant on-screen pixel size, dividing by `Camera.Zoom`) instead of `Absolute` (scales with the rest of the shape, like Width/Height) — MonoGame/raylib never set this explicitly so they use the enum default (`Absolute`). This was introduced with Skia's two-slot fill+stroke composition (#2790/#2814); a few existing tests already had to override it back to `Absolute` to get deterministic pixel measurements, which was the tell.

Fixed both defaults to `Absolute`. New pinning tests in `RectangleRuntimeTests`/`CircleRuntimeTests`. Full SkiaGum.Tests (428) still green, samples rebuild clean.

PR #4376, opened 2026-08-07
Compose oversample compensation into Raylib BBCode default/[FontScale] runs

### LINKED ISSUE #4365: Raylib font oversampling doesn't compensate inline BBCode runs (FontScale/FontSize tags)

vchelaru:
Closes #4365

`ResolveRunFont` (Raylib `Text.cs`) resolved a styled BBCode run's font/size from `_font.BaseSize`, which IS the oversampled raster once `RegenerateOversampledFont` swaps it in, without ever composing `OversampleCompensationScale`. A run with no tag, or an explicit `[FontScale=v]` tag, therefore ballooned in size with the oversample ratio instead of staying pinned to native size — unlike plain text (`DrawPlainLine`/`EffectiveFontScale`), and unlike MonoGame/KNI/FNA where this already composed correctly.

Fix composes `OversampleCompensationScale` into `ResolveRunFont`'s default and `[FontScale]`-tag branches, mirroring `EffectiveFontScale`/`DrawPlainLine`, and updates `DrawStyledLine`'s baseline math plus `GetLineLayoutScale`/`GetLineTopOffsets` (multi-line vertical stacking) to compare against the same now-compensated scale so height/baseline math stays consistent.

`[FontSize=...]`/`[IsBold=...]`/`[IsItalic=...]`/`[OutlineThickness=...]` swap-font runs are intentionally left uncompensated — same scope cut #4330 made, since composing compensation into that resolution risks destabilizing the line-height/baseline math those runs depend on. Docs updated to narrow the documented limitation accordingly.

## Test plan
- [x] `RegenerateOversampledFont_BBCodeColorOnlyRun_WidthDoesNotDriftFromNative` (new, red before fix — width inflated ~2.5x, matching the oversample ratio)
- [x] `RegenerateOversampledFont_BBCodeFontScaleTagRun_WidthDoesNotDriftFromNative` (new, same red/green)
- [x] Full `RaylibGum.Tests` suite green (589/589)


vchelaru:
## Update

Manual testing surfaced two more real bugs in the same area, fixed here too:

- **Line spacing grew under oversampling.** `LineHeightInPixels` was baked from the live (possibly oversampled) Font's raw metrics with no compensation. Added `EffectiveLineHeightInPixels`, used in `Render()`'s plain-line advance and `GetLineTopOffsets()`.
- **BBCode word-wrap changed depending on whether oversampling was on.** Wrap/pre-render measurement (`ResolveRunFont`) measured against the live Font, a different hinted raster than pre-oversampling. `ResolveRunFont` now takes an explicit base-font/compensation pair; the measurement call site uses the stable `EffectiveMeasurementFont` with no compensation, matching how plain text already avoids this.
- **Sample fix:** the raylib gallery's Forms controls (buttons/checkboxes/radio buttons) visibly shrank once oversampling engaged. Traced to `InitializeStyling` assigning fonts via a raw `Font` struct, which bypassed the pipeline `RegenerateOversampledFont` uses to pin its native reference font — pinned vs. generated ended up as different assets, skewing the compensation ratio. Switched to `CustomFontFile`+`FontSize`. (Same class of gap likely also affects MonoGame's mirror of this code — not filed yet, flagging here.)

New tests: `RegenerateOversampledFont_PlainMultiLineText_LineSpacingDoesNotDriftFromNative`, `MeasureString_BBCodeWord_DoesNotDriftFromNative_WhenOversamplingActive`, `RegenerateOversampledFont_BBCodeText_WrappingDoesNotChangeFromNative`. Full suite: 592/592 green.

PR #4398, opened 2026-08-09
Fix TextBox scrolling horizontally when single-line-mode text renders on multiple lines

### LINKED ISSUE #4393: Gum shouldn't have incorrect character spacing causing characters to disappear in multiline text
https://github.com/user-attachments/assets/4a765aef-f923-4bd9-9e65-a99591a1af80

vchelaru:
A TextBox configured as single-line (`NoWrap`, no `AcceptsReturn`) still renders multiple visual lines when the assigned text contains a `\n`, or when the visual's text instance is sized relative to its parent and wraps. Caret placement, hit-testing and selection all keyed off the configuration flag rather than the layout, so they measured the whole string as one line — the caret for an index on line 2 landed hundreds of pixels past the right edge, and the horizontal scroll dragged the text left to chase it. That's the jump in the video: `TextInstance.X` went `4` → `-258`.

Those three now gate on the rendered line count (`IsRenderedAsSingleLine`). Vertical scrolling still gates on configuration, preserving the invariant that single-line boxes never shift the text vertically.

Also adds `TextBoxWrappingTests` — wrap-mode horizontal behavior, the state↔wrapping contract, and caret/hit-test agreement on wrapped lines had no coverage at all. Each of the three changed branches goes red without the fix.

Closes #4393

### DIFF
--- a/MonoGameGum.Tests/Forms/TextBoxWrappingTests.cs
+++ b/MonoGameGum.Tests/Forms/TextBoxWrappingTests.cs
@@ -1,0 +1,312 @@
+using Gum.Forms.Controls;
+using Gum.Forms.DefaultVisuals;
+using Gum.Wireframe;
+using Shouldly;
+using System.Linq;
+using Xunit;
+
+namespace MonoGameGum.Tests.Forms;
+
+/// <summary>
+/// Coverage for the interaction between wrapping and horizontal scrolling. A TextBox whose text
+/// wraps must never scroll horizontally — every visual line already fits the container, so a
+/// leftward shift only pushes characters out of the clipped region. See issue #4393.
+/// </summary>
+public class TextBoxWrappingTests : BaseTestClass
+{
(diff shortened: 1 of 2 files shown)


vchelaru:
Reproduced against Airpig with a file-based probe. The real trigger is different from the first commit's:

Airpig's TextBox component has no `MultiNoWrap` state in its `LineModeCategory`, so `AcceptsReturn = true` with `TextWrapping = NoWrap` falls back to the `Multi` state — which sizes the text relative to its parent and therefore **wraps** — while horizontal scrolling stays enabled because the property still says `NoWrap`. On the keystroke that wraps a line the caret is measured at the end of that line including the trailing space (caret X 123 in a 120px box), so the text lurches left.

Latest commit gates the horizontal branch on the text component's width units as well as `TextWrapping`. The caret-past-the-edge half is filed separately as #4399.

PR #4440, opened 2026-08-11
Fix DialogBox.Dismiss raising FinishedShowing before cleanup

ultimateownsz:
## Summary

\DialogBox.Dismiss()\ cleared the dialog's \Pages\ and dropped keyboard focus **after** raising \FinishedShowing\:

\\\csharp
public void Dismiss()
{
    this.IsVisible = false;
    ...
    PageAdvanced?.Invoke(this, null);
    FinishedShowing?.Invoke(this, null);  // fired BEFORE cleanup
    this.Pages.Clear();
    IsFocused = false;
}
\\\

A handler that re-shows the dialog from \FinishedShowing\ (the natural implementation of chained/sequential dialogs) had its freshly-shown pages and focus wiped by the tail of \Dismiss()\. Result: the next dialog box is visible but stuck — no pages, no input focus.

## Fix

Move \Pages.Clear()\ and \IsFocused = false\ before \FinishedShowing\ fires, so the dialog is fully dismissed by the time the event is raised. This also matches the \ShowAsync\ path, which only fires the event after the show loop has fully ended (its cleanup happens before the event too).

## Tests

Two new tests in \MonoGameGum.Tests/Forms/DialogBoxTests.cs\:
- \Dismiss_ShouldClearPagesAndDropFocus_BeforeFinishedShowingFires\ — pins the ordering contract.
- \FinishedShowingHandler_ReShow_ShouldSurviveDismissCleanup\ — the real-world chained-dialog scenario: re-showing a 2-page dialog from the handler must keep its remaining page and focus.

Both fail on the previous ordering and pass with the fix. Full \MonoGameGum.Tests\ suite passes (2174).

### DIFF
--- a/MonoGameGum.Tests/Forms/DialogBoxTests.cs
+++ b/MonoGameGum.Tests/Forms/DialogBoxTests.cs
@@ -259,4 +259,51 @@ public class DialogBoxTests : BaseTestClass
         // Three short entries -> three pages, first popped -> two remaining.
         dialogBox.PagesRemaining.ShouldBe(2);
     }
+
+    [Fact]
+    public void Dismiss_ShouldClearPagesAndDropFocus_BeforeFinishedShowingFires()
+    {
+        var (dialogBox, visual) = CreateDialogBox();
+        dialogBox.LettersPerSecond = 0;
+        dialogBox.IsFocused = true;
+
+        dialogBox.Show(new[] { "one", "two", "three" });
+        dialogBox.PagesRemaining.ShouldBeGreaterThan(0); // sanity: pages queued
+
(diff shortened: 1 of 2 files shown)


vchelaru (approved):
Verified the Pages-null concern doesn't apply — `Pages` is a private `List<T>`, never exposed publicly, never null. Reordering matches the existing async-path (`StartShowAllPagesLoop`) behavior. Confirmed both `Pages.Clear()` and `IsFocused = false` need to move together (traced through `Show()`'s own `Pages.Clear()`+repopulate) — the two new tests correctly pin that, verified they'd fail on a partial fix. LGTM.

PR #4456, opened 2026-08-14
Copy and paste a whole variable category from the Variables tab context menu

### LINKED ISSUE #4455: Variables tab: right-click a category header to copy a group of variable values and paste it onto other objects
## Problem

There is no way to copy a whole group of variables from one object to another. If a mockup has the font settings you want (Font Source, Font, Font Size, Outline Thickness, Is Bold, Use Font Smoothing, Font Scale), matching a real Text object to it means reading each row on one object, selecting the other, typing the value, and going back. That is one round trip per variable.

The Variables tab already groups variables into exactly the right units for this: Text, Font, Flip and Rotation, Position, Dimensions. Those groups are `MemberCategory` objects, so "copy this group's values" has an obvious anchor.

## Suggested improvement

Right-click a category header in the Variables tab and get Copy / Paste for the whole group.

Copy captures the effective value of every row in the category (what is displayed, whether it is set locally or inherited). Paste writes them onto the current selection, and because the grid already routes a multi-selection through `MultiSelectInstanceMember`, pasting onto several selected instances at once applies to all of them under a single undo. So the loop becomes: copy once from the mockup, select every target, paste once.

Paste skips rather than fails on rows the target cannot take:

- variable not present on the target (pasting Font onto a Sprite)
- read-only rows (locked instances)
- rows driven by a `VariableReferences` entry, which paste should not silently shadow

Design notes on the two variants considered and rejected for a first pass:

- **Copy To...** (stay on the source, pick the destination from a dialog) does not fit the existing write path. `VariableGridEntry.SetValue` writes to `_selectedState.SelectedStateSave`, so writing into an element that is not selected means bypassing the grid write path and hand-rolling refresh and undo.
- **Copy From...** (sit on the target, pull values from a source picked in a dialog) does fit, because reading from a non-selected element is safe and the write still lands on the selected element. This is a good follow-up on top of the same payload plus applier, and only adds a source picker.

## Source

- `Tools/Gum.Presentation/Plugins/InternalPlugins/VariableGrid/CategorySortAndColorLogic.cs:23` defines the category names being grouped.
- `Tools/Gum.Presentation/Plugins/InternalPlugins/VariableGrid/VariableGridEntry.cs:558` `SetValue` is the per-row write (state write plus notify/refresh cascade); the read side is `Value` / `GetRootVariableSave()`.
- `Gum/Plugins/InternalPlugins/VariableGrid/CompositeMemberLogic.cs:263` shows the pattern for batching several writes into one undo (`_undoManager.RequestLock()`).
- `WpfDataUi/DataTypes/MemberCategory.cs` has no context-menu hook; `InstanceMember.ContextMenuEvents` (`WpfDataUi/DataTypes/InstanceMember.cs:58`) is the shape to mirror.
- The category Expander is declared in two templates and neither has a `ContextMenu`: `WpfDataUi/Themes/Generic.xaml:33` (default) and `Gum/Themes/Frb.Styles.Defaults.xaml:1702` (`DataUiGrid.VariableGrid`, the one the Variables tab uses).

The gather/apply logic can live headless in `Gum.Presentation` and be unit tested without WPF.

## Reach

Any user matching one object's appearance to another, which is the normal way people work from a mockup. It gets worse the more variables the group has, and Font is the largest group in practice.

kaltinril:
Closes #4455.

Right-click a category header in the Variables tab (Font, Text, Position, ...) and you get **Copy Values** and **Paste Values**. Matching one object to another no longer means reading a value off one object, selecting the other, typing it, and going back for the next one.

Copy captures the effective value of every row in the category (what is displayed, whether authored on the selected state or inherited). Paste writes them onto the current selection, matching by variable name, recorded as a single undo. Because the grid already routes a multi-selection through `MultiSelectInstanceMember`, selecting several objects and pasting once applies to all of them, so the loop is: copy once from the mockup, select every target, paste once.

Paste is deliberately lenient. It skips rather than fails on a variable the target does not have (pasting Font onto a Sprite), read-only rows (locked instances), rows driven by a `VariableReferences` assignment, values whose type the target row does not accept, and values the target already shows. The outcome is reported to the output window: `Pasted 6 value(s) into Font. Skipped FontScale, OutlineThickness`.

Never copied, regardless of category: `Name`, `BaseType`, `Locked`, `Parent`, `VariableReferences`. These identify or gate an object rather than describe it, and a pasted parent name is dangling on another element. List-valued variables are skipped too, since the value is held by reference and source and target would otherwise share one instance.

## Shape

The gather/apply logic is headless in `Gum.Presentation` (`VariableCategoryCopyPasteService`) and unit tested there without WPF. The Variables tab's WPF rows are adapted to it through `IVariableCategoryRow`.

Rows are taken exactly as the grid presents them, including composite rows (the color swatch) and multi-select wrappers. An earlier revision expanded composites into their Red/Green/Blue channels; that was wrong, because the multi-select wrapper sits *above* the composite, so the same row was named Red/Green with one object selected and Color with several, and a copy made in one mode silently did not apply in the other. Writing through the row as-is also preserves each wrapper's own behavior: the composite skips unchanged channels, and the multi-select wrapper fans the value out and refreshes afterwards.

`MemberCategory` gained a `ContextMenuItems` collection (`MemberCategoryContextMenuItem` is its own `ICommand`, so the menu is built purely by binding). The category header template in `Frb.Styles.Defaults.xaml` hosts the menu; categories that add no items show nothing at all, so other `DataUiGrid` consumers are unaffected.

## Design notes on the variants not built

- **Copy To...** (stay on the source, pick the destination from a dialog) does not fit the existing write path. `VariableGridEntry.SetValue` writes to `_selectedState.SelectedStateSave`, so writing into an element that is not selected means bypassing the grid write path and hand-rolling refresh and undo.
- **Copy From...** (sit on the target, pull from a source picked in a dialog) does fit, since reading from a non-selected element is safe and the write still lands on the selected element. It reuses this PR's payload and applier and only adds a source picker, so it is a natural follow-up if the two-gesture flow feels like one step too many.

## Testing

- `Tests/Gum.Presentation.Tests/VariableCategoryCopyPasteServiceTests.cs` covers apply/skip matching, read-only and reference-assigned rows, the type guard, the never-copied set, the already-matches case, and that one undo lock is held across every write.
- `Tool/Tests/GumToolUnitTests/VariableGrid/VariableCategoryRowAdapterTests.cs` pins the composite naming being identical in single- and multi-select, and write-through for composite and multi-select rows.
- `Gum.Presentation.Tests` 915 passed, `GumToolUnitTests` 1269 passed, `GumFull.sln` builds with no new warnings.
- Not unit tested: the `PropertyGridManager` glue that hangs the menu items on the live grid, and the XAML. Those need a manual pass in the running tool.

## Manual test steps

1. Open a project, select a Text object with fonts set, right-click the **Font** header, Copy Values.
2. Select a different Text object, right-click **Font**, Paste Values. The font settings should match and one Ctrl+Z should undo the whole group.
3. Select several Text objects at once and paste. All of them should update, again as one undo.
4. Right-click **Font** on a Sprite and paste. Nothing should change, and the output window should say everything was skipped.
5. Right-click a category header on a locked instance and paste. Values should be skipped rather than written.

## Boyscout

`MemberCategory`'s two constructors were chained and `HandleMembersChanged`'s duplicated switch arms merged with a null guard, clearing four pre-existing nullable warnings in the file.

### DIFF
--- a/.claude/skills/gum-tool-variable-grid/SKILL.md
+++ b/.claude/skills/gum-tool-variable-grid/SKILL.md
@@ -107,6 +107,22 @@ Landmine: `CompositeInstanceMember.HandleCustomSet` writes **every** channel on 
 
 ---
 
+## Category Copy/Paste (whole-group values)
+
+Right-clicking a category header offers Copy Values / Paste Values (#4455). The gather/apply logic is headless: `VariableCategoryCopyPasteService` in `Tools/Gum.Presentation/.../VariableGrid/`, fed by `IVariableCategoryRow` rows that `VariableCategoryRowAdapter` (in `Gum/.../VariableGrid/`) wraps around the live WPF `InstanceMember`s. Paste only applies to the category the values were copied from, records as one undo (single `RequestLock` across all writes), and skips rather than fails: absent variables, read-only/reference-assigned rows, type mismatches (numeric mismatches convert), and state names the target doesn't declare.
+
+Category-header menus are `MemberCategory.ContextMenuItems` (`MemberCategoryContextMenuItem` is its own `ICommand`), rendered by the header template in `Frb.Styles.Defaults.xaml`. Landmines:
+
+- **`DataUiGrid.SetMultipleCategoryLists` copies only `Name` and `HeaderColor`** onto the fresh `MemberCategory` objects it creates for multi-select — any other category-level state (`ContextMenuItems`, `HideHeader`, `Width`) is silently dropped. `PropertyGridManager.RefreshCategoryContextMenus` re-populates after every refresh (guarded to only fill empty categories) for exactly this reason.
+- **A `ContextMenu` on a shared template reaches every grid.** The header template serves all `DataUiGrid`s (Code tab, project properties, behaviors), and even a visually-collapsed empty menu still opens and takes mouse capture, eating the next click. Suppress with `ContextMenuService.IsEnabled=False` (via DataTrigger on item count), not `Visibility`.
+- **`Type.IsInstanceOfType` is always false against `Nullable<T>`** for a boxed value — variables declared `int?`/`float?` (`MaxLettersToShow`, `MinWidth`, ...) need `Nullable.GetUnderlyingType` before any reflection type gate.
+- **A state name is just a string** — any automated write to a state row (`State`, `<Category>State`) must validate against the row's `CustomOptions`/available states or it saves a dangling state name that renders as a blank combo.
+- **Write units before values.** Setting `XUnits`/`WidthUnits`/... can convert the target's current `X`/`Width` ("convert variables on unit type change"), so a value written before its unit gets converted away.
+- **`MultiSelectInstanceMember.Value` returns null when the wrapped instances disagree** (indeterminate) — indistinguishable from "unset" by value alone; check `IsIndeterminate` before treating null as absent.
+- Right-clicking inside a value text field pops an empty white box (#4457, cause undetermined — candidates: a displayer's declared-but-never-populated `<ContextMenu />`, or the native text-editing menu unthemed). Pre-existing on main; don't chase it as a regression of grid work.
+
+---
+
 ## Refresh Trigger Flow
 
 ```
--- a/.claude/skills/gum-tool-variable-grid/SKILL.md
+++ b/.claude/skills/gum-tool-variable-grid/SKILL.md
@@ -119,7 +119,7 @@ Category-header menus are `MemberCategory.ContextMenuItems` (`MemberCategoryCont
 - **A state name is just a string** — any automated write to a state row (`State`, `<Category>State`) must validate against the row's `CustomOptions`/available states or it saves a dangling state name that renders as a blank combo.
 - **Write units before values.** Setting `XUnits`/`WidthUnits`/... can convert the target's current `X`/`Width` ("convert variables on unit type change"), so a value written before its unit gets converted away.
 - **`MultiSelectInstanceMember.Value` returns null when the wrapped instances disagree** (indeterminate) — indistinguishable from "unset" by value alone; check `IsIndeterminate` before treating null as absent.
-- Right-clicking inside a value text field pops an empty white box (#4457, cause undetermined — candidates: a displayer's declared-but-never-populated `<ContextMenu />`, or the native text-editing menu unthemed). Pre-existing on main; don't chase it as a regression of grid work.
+- Right-clicking inside a value text field can pop an empty-looking white box (#4457): the native text-editing menu, unthemed — disabled Cut/Copy/Paste entries render invisibly faint, so with no selection and an empty clipboard the menu looks blank. Pre-existing on main; don't chase it as a regression of grid work.
 
 ---
 
--- a/Gum/Plugins/InternalPlugins/VariableGrid/IVariableCategoryRowAdapter.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/IVariableCategoryRowAdapter.cs
@@ -1,0 +1,14 @@
+using System.Collections.Generic;
+using WpfDataUi.DataTypes;
+
+namespace Gum.Plugins.InternalPlugins.VariableGrid;
+
+/// <summary>
+/// Adapts the Variables tab's WPF rows to <see cref="IVariableCategoryRow"/> so the headless
+/// <see cref="IVariableCategoryCopyPasteService"/> can read and write them without referencing WPF.
+/// </summary>
+public interface IVariableCategoryRowAdapter
+{
+    /// <summary>Wraps each member of a category as a copy/paste row.</summary>
+    List<IVariableCategoryRow> CreateRows(IEnumerable<InstanceMember> members);
+}
--- a/Gum/Plugins/InternalPlugins/VariableGrid/PropertyGridManager.cs
+++ b/Gum/Plugins/InternalPlugins/VariableGrid/PropertyGridManager.cs
@@ -87,7 +87,8 @@ public partial class PropertyGridManager : IBehaviorVariablePropertyGridSink
     private CompositeMemberLogic _compositeMemberLogic;
     private StateSaveCategoryDisplayer _stateSaveCategoryDisplayer;
     private BehaviorShowingLogic _behaviorShowingLogic;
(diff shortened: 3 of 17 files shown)


kaltinril:
Manually tested in the running tool. Two things worth recording for review.

**Verified working:** a pasted group undoes as one action. That is now pinned by `VariableCategoryCopyPasteUndoTests`, which drives the real `UndoManager` rather than a mock, since the point of the lock the service takes is how it interacts with undo recording. Confirmed the test fails when the lock is removed (a single undo then rolls back only the last variable and leaves the rest of the group applied).

**Scoping confirmed correct, with a caveat worth a second opinion.** An initial test looked like paste was writing far more than the chosen category. It was not. Instrumented logging showed a copy of `States and Visibility` capturing exactly its three rows and a paste applying exactly those three:

```
--- COPY on category 'States and Visibility' ---
    grid categories: General(4) | Position(6) | Dimensions(8) | Text(7) | Font(9)
                     | Flip and Rotation(1) | States and Visibility(3) | Parent(2)
                     | Rendering(2) | Behavior(2) | References(1)
    row 'Visible'            value=True
    row 'ColorCategoryState' value=White
    row 'StyleCategoryState' value=Large
    captured: Visible, ColorCategoryState, StyleCategoryState
```

The on-disk diff of the component agreed: two variables written, and the target's `Text` variable untouched.

What made it *look* like everything changed is `StyleCategoryState`. Assigning a style state re-styles the object wholesale, and the grid then renders the values that state supplies, so the Text and Font rows visibly change while nothing is written to them.

So the mechanism is existing Gum behavior, not the paste. The open question is whether a categorized state variable belongs in a bulk category copy at all: it is one row whose blast radius is an entire named style, which is a different kind of thing from "make the font match". This PR currently keeps state variables copyable. Excluding them is a one-line change to `NeverCopiedVariableNames` if you would rather they were left out.

kaltinril:
Ran a multi-angle review pass over the branch (correctness, reuse, efficiency, conventions, removed-behavior, cross-file tracing). Commit c5f8b37 lands the accepted findings; this comment records what was found, what was fixed, and the calls that are yours to overrule.

## Fixed (each with a pinning test where testable)

- **Nullable-declared variables never pasted.** `MaxLettersToShow` (`int?`), `MinWidth`/`MaxWidth`/`MinHeight`/`MaxHeight` (`float?`): reflection's `IsInstanceOfType` is always false for a boxed value against `Nullable<T>`, so the type gate skipped them every time. Now compares against the underlying type.
- **Dangling state names.** A state row's value is just a string, so pasting `States and Visibility` across element types would save a state name the target does not declare. State-combo rows now validate the value against the row's available states; same-element pastes still work (that path was live-tested).
- **Unit/value ordering.** With "convert variables on unit type change" on, writing `X` before `XUnits` let the unit change convert the just-pasted `X`. Units are now applied first.
- **Multi-select copy silently dropped disagreeing variables** (indeterminate reads as null). They are now reported in the copy output: `Not copied (differs across selection): FontSize`.
- **Every DataUiGrid category header in the app** (Code tab, project properties, behaviors) got an invisible context menu that opened on right-click and swallowed the next click. Empty categories now disable the menu via `ContextMenuService.IsEnabled`.
- **Honest paste reporting.** Values the target already showed were counted as "Pasted" and triggered a full grid rebuild with nothing written. The result now separates applied / already matched / skipped, and the rebuild is gated on real writes.
- **`DefaultChildContainer`** added to the never-copied set (cross-element paste re-points the target's default slot at a child it does not have), plus **numeric coercion** (int 36 onto a float row converts instead of skipping), and perf work: `ObjectFinder` cache around copy/paste, single effective-value read per row, no per-row grid-wide re-read (`CallAfterSetByUi` dropped - nothing subscribes `AfterSetByUi`), menus built once per category object.

## Considered and declined - flagging for your judgment

- **Per-variable write pipeline.** Each pasted variable still runs the full set pipeline (auto-save, plugin broadcast, and for `VariablesRequiringRefresh` names like `Font` a mid-paste grid rebuild). Cost equals manually editing N fields, so it is correct but not batched. Batching would mean the `IsCallingRefresh`-style suppression plus one save/notify pass at the end - a deeper change to a live-verified path than I wanted to make without in-tool testing. Related known edge: rows that only *appear* because of a pasted gate variable (paste `UseCustomFont=true` onto a target without it: `CustomFontFile` is reported Skipped because the row did not exist when the paste started).
- **Multi-select rejections are swallowed.** `MultiSelectInstanceMember`/`CompositeInstanceMember` fan out writes without aggregating inner failures, so a rejected inner write still reports as applied on those paths. Pre-existing wrapper semantics; single-select reports honestly.
- **`NeverCopiedVariableNames` is a hardcoded list** - the repo's fourth such (with `CopyPasteLogic`, `GumFileSerializer`, `VariableReferenceLogic`), and they already disagree (e.g. `IsSlot`). The data-driven home would be a flag on the variable definition (near `CanOnlyBeSetInDefaultState`/`ExcludeFromInstances`), but that is a data-model decision worth making once for all four call sites rather than in this PR.
- **Hand-rolled `ICommand` instead of `RelayCommand`**: WpfDataUi has no CommunityToolkit dependency, and `CommandManager.RequerySuggested` auto-requery is what keeps "Paste Values" enabling itself after the first copy without event plumbing.
- **"Already matched" does not author explicitly.** If the target *inherits* the copied value, paste leaves it inherited (reported as matched, not written). Deliberate: no-op writes should not convert inherited values to authored ones. The counter-argument is that the user wanted a hard stamp; flip `TryApply`'s equality branch if you prefer that.

Suites: Gum.Presentation.Tests 921, GumToolUnitTests 1270, zero warnings in touched files. The XAML `ContextMenuService.IsEnabled` change and the state-name guard's accept path are the two spots that most deserve a quick in-tool check.

kaltinril:
Live-verified on the final build (670d035) against a real project:

- Paste Values is greyed on every category except the one the values were copied from; pasting Font values onto another Text works and reverts with a single Ctrl+Z.
- No right-click regressions on other tabs, grids, or the tree view; category headers without menu items do nothing on right-click.
- Same-element state paste and multi-select paste were exercised in earlier rounds tonight.

One pre-existing issue surfaced during testing, confirmed present on main and unrelated to this branch: right-clicking inside a variable value text field (e.g. Base Type's editable combo) shows WPF's native Cut/Copy/Paste menu rendered white-on-white, so it looks like an empty box. Will file separately.

PR #4475, opened 2026-08-19
Get the Gum tool compiling and running from source on Linux (via Wine)

profexorgeek:
## Summary

Gum's tool (`GumFull.sln`) previously couldn't even compile on Linux/Mac, and there was no supported way to build-and-run your own source changes there. This gets both working, using Wine as the runtime (native cross-platform WPF is a separate, much larger, unscoped effort - see `Direction/open-questions.md`).

- `GumFull.sln` now compiles cleanly on Linux (0 errors).
- A locally-built Gum now launches and runs correctly through Wine, including loading and editing real projects.
- **Not a replacement for Windows development** - see Known Limitations below.

## What changed

1. **`Directory.Build.props`** - opts into `EnableWindowsTargeting` so the SDK will compile `net8.0-windows` tool/plugin projects on non-Windows OSes (`NETSDK1100`). Compile-only; no-op on Windows.
2. **10 plugin `.csproj` files** - their post-build steps were raw Windows `cmd.exe` batch (`xcopy`, `IF EXIST`/`rd /s /q`, `md`), which doesn't exist on Linux (`MSB3073`). Rewrote them as portable MSBuild `Copy`/`RemoveDir` tasks - same behavior, works identically on Windows too (no more `cmd.exe` dependency there either). Verified the resulting `Plugins/`/`Content/FormsThemes/` output matches the original batch output file-for-file.
3. **`PluginManager.cs`** - a plugin-assembly load failure was silently swallowed (`catch { // todo - report the error }`). Now reported via `IOutputManager.AddError`. Found while ruling out causes of the Linux rendering bug (see below); kept as a standalone fix since silent plugin failures are bad on any platform.
4. **`Program.cs`** - added `AppDomain.UnhandledException`/WinForms `Application.ThreadException`/`TaskScheduler.UnobservedTaskException`/WPF `DispatcherUnhandledException` handlers that log to `Console.Error`. None of these were wired up before, so an exception off the main thread could vanish completely. Also found while debugging the rendering issue (all four stayed silent, which is what proved it *wasn't* a swallowed exception) and kept for the same reason as (3).
5. **`run_gum_linux.sh`** (new) - launches a locally-built Gum through an existing Wine prefix. Complements `setup_gum_linux.sh` (which installs the official *released* build) - this one is for testing your own source changes.

## What we verified along the way

- **Wine version floor**: `setup_gum_linux.sh` already required Wine ≥ 10 (added in #1524) with no documented reason beyond "what we happened to test with." We tested lowering it to 9 against a real machine (Linux Mint 22.3's packaged `wine-9.0~repack-4build3`) and it fails with a genuine, reproducible crash: `FileLoadException`/`Incorrect alignment (0x8007046C)` loading `System.Private.CoreLib` region during CoreCLR init. This matches a publicly-documented Wine defect (see sources below) fixed in a later 9.x point release. Wine 11.0 (current WineHQ stable) works correctly. **Left the floor at 10** - confirmed necessary, not just cautious.
- **Getting `GumFull.sln` to compile**: `NETSDK1100` (fixed by `Directory.Build.props`) → `MSB3073` from the batch post-builds (fixed by the MSBuild task rewrite) → clean build.
- **Getting the *official released* build running under Wine**: hit and resolved, in order: a `DOTNET_ROOT`/PATH leak pointing hostfxr at the Linux SDK install (already handled by the existing `~/bin/gum` wrapper's `unset`, but I re-broke it once by testing without going through the wrapper), a Wine-Mono "package not found" install-prompt dialog blocking `wineboot -u` (harmless - Gum uses a real installed .NET 8 runtime, not Wine's Mono/.NET-Framework shim - just needs to be clicked through or the prefix's `DllOverrides\mscoree` left at its default), and the Wine 9→11 alignment bug above. Confirmed fully working: loads real projects, editor functions correctly.
- **Getting a *locally-built* Gum running the same way**: this took much longer. A `dotnet build`-produced `Gum.dll`, launched via `wine dotnet Gum.dll`, ran with zero errors/exceptions anywhere (verified via the new global exception hooks, `IOutputManager` output, and Wine debug tracing) but its render viewport (`EditorTabPlugin_XNA`, the KNI-based canvas) never initialized - blank canvas, no D3D/DXGI activity in Wine's trace at all. Ruled out, with evidence: the MSBuild post-build rewrite (byte-identical plugin deployment vs. the official build), Debug-vs-Release configuration, a duplicate-assembly type-identity conflict, and a swallowed exception (nothing in the whole chain of exception hooks ever fired). The actual cause: `dotnet build` never produces a native `Gum.exe` apphost, only the framework-dependent `.dll` - and under Wine specifically, launching via `dotnet Gum.dll` (the muxer) instead of a real apphost exe breaks KNI's render initialization with no error of any kind. A full `dotnet publish -r win-x64` apphost fixes it but is slow and produces a 100MB+ bundle; a lean `dotnet build -r win-x64 --self-contained false -p:UseAppHost=true` (263KB) fixes it just as well. `run_gum_linux.sh` uses the latter.

## Known limitations

- **Debugging (breakpoints, F5-with-debugger) does not work through Wine.** Researched this before starting: WPF's rendering layer requires Wine no matter what, but IDE debuggers (Rider, Visual Studio, VS Code) attach via a protocol that assumes the debugger and the target CoreCLR share a native OS ABI - a Wine-hosted process doesn't satisfy that. A dedicated "porting WPF to Linux via Wine" guide reports the same: VS/VS Code remote debugging attempts failed outright; the one documented working exception is `dnSpy` running inside the same Wine prefix (IL-level, not integrated source debugging). Rider isn't documented as a working exception either, and its debugger architecture is the same general shape. **You can build and run from source; you cannot attach a debugger.**
- This is **not an officially supported development path** - see the "Documentation notes" section below for the disclaimer we think a docs page should carry.

## Documentation notes (for a future docs PR)

Per repo convention, doc content isn't part of a code PR - but here are detailed notes for whoever picks up writing this, aimed at a new page (suggest `docs/gum-tool/setup/running-from-source-linux.md`, as a sibling to the existing Windows-only `running-from-source.md`, cross-linked from it and added to `SUMMARY.md`):

**Framing / disclaimer** (put this near the top):
- This is **not an officially supported path**. It works, but it's Wine-based, has a real known limitation (no debugger), and hasn't been tested as broadly as the Windows workflow. Treat it as "works for us, your mileage may vary," not a guarantee.

**Prerequisites**:
- Wine **10 or newer** is required, not just "any recent Wine." We confirmed Wine 9.0 (at least the Linux Mint-packaged build) fails outright with a CoreCLR loading crash (`0x8007046C`, "incorrect alignment") - a real, documented Wine regression in early 9.0 builds, not a soft recommendation. Point to the WineHQ-signed-repo install steps in `setup_gum_linux.sh` rather than a distro's default package if the default is older than 10 (Mint's own Software Manager, for example, currently ships 9.0).
- The standard Linux dev machine setup: run `setup_gum_linux.sh` once to get Wine + the correct fonts + the .NET 8 desktop runtime installed into a Wine prefix (default `~/.wine_gum_dotnet8`).

**Two scripts, two different purposes - document them as clearly separate, not variations of each other**:
- `setup_gum_linux.sh`: installs Wine/winetricks/.NET 8 runtime, downloads the latest **officially released** `gum.zip`, and sets up a `~/bin/gum` launcher. This is for people who just want to *run* Gum on Linux, not build it. Also handles `gum upgrade`/`gum dxvk`/`gum d3d`.
- `run_gum_linux.sh` (new, this PR): for **Gum contributors**. Assumes you've already run `setup_gum_linux.sh` once (for the Wine prefix) and separately run `dotnet build GumFull.sln` to build your own checkout. Launches *that* build - not the downloaded release - through the existing prefix. This is the "test my source changes" loop, analogous to pressing F5 in Visual Studio on Windows, minus the debugger (see below).
- Worth an explicit note on *why* `run_gum_linux.sh` builds a native apphost exe instead of just running the built `.dll` directly - that's the least obvious part of the whole setup, and the reasoning is in the commit message / PR description above if the doc author wants the full technical detail.

**Known limitation - no debugging**:
- State plainly: you can build and run Gum from source on Linux, but you cannot attach a debugger (no breakpoints, no step-through) - not in Rider, not in any IDE. This is a Wine/CoreCLR-debugging-protocol limitation, not something specific to Gum or fixable by us. If someone needs to actually debug tool code, they need a Windows machine or VM.
- Could mention `dnSpy`-inside-Wine as the one documented (non-integrated) alternative for advanced users, with a caveat that it's IL-level decompiler-based debugging, not source-level.

**Possible IDE integration section** (Rider specifically, since that's what was tested):
- Rider's own built-in "Run"/"Debug" for the Gum project won't work directly on Linux (same underlying reason `run_gum_linux.sh` exists - the built `.dll` can't just be executed, and even the real exe needs Wine, not a native Linux process launch).
- The practical setup is a Rider **Shell Script** run configuration: "Before launch" = Build (`GumFull.sln`), then run `run_gum_linux.sh`. This gives a one-click "compile + launch" loop equivalent to F5 *without* the debugger. Worth documenting as the recommended IDE-integration pattern, explicitly labeled as Run-only, not Debug.


profexorgeek:
Functionality demo:

[cinnamon-2026-08-18T195301-0600.webm](https://github.com/user-attachments/assets/f98d2bd8-d04c-4abf-a6dd-96fd269d049b)

vchelaru:
Impressive stuff!!

Issue #4477, opened 2026-08-19
Support Add color operation from .achx/.achj animation chains


vchelaru:
## Problem
FlatRedBall2's `AnimationChain.Common` recently added per-frame color support to the `.achx`/`.achj` format: nullable `Red`/`Green`/`Blue`/`Alpha` (int, 0-255) plus a `ColorOperation` enum (`Multiply`/`Add`), authored in the FlatRedBall Animation Editor. `Multiply` and `Alpha` are tracked separately in #4490 — they need no rendering changes since `Sprite`/`NineSlice` already tint via a standard SpriteBatch multiply. This issue is scoped to `Add`, which does need new rendering work.

Reference: `FlatRedBall2/src/AnimationChain.Common/ColorOperation.cs`, `AnimationChainListSave.cs` (frame read/write), and the runtime application in `FlatRedBall2/src/Rendering/SpriteFrameColor.cs` / `AnimationChain.MonoGame/Runtime/AnimationFrameColor.cs`.

## Suggested improvement
`Add` can't be expressed as a color multiply — `SpriteBatch` (and Gum's renderers) only multiply per-pixel color, they can't offset it. FRB2 routes `Add` through a dedicated pixel shader (`AddColorEffect`/`WorldSpaceAddColorBatch`). Gum needs an equivalent shader/effect path per backend (MonoGame/KNI/FNA, raylib, Skia) for frames whose `ColorOperation` is `Add` — this is the main scoping work here, not the data model (that part lands with #4490, which adds the `ColorOperation` enum value itself).

## Reach
Any `.achx`/`.achj` file authored in the current FlatRedBall Animation Editor that uses per-frame `Add` color (flash/hit effects) will have that data silently dropped when loaded into Gum today.

vchelaru:
Closing as speculative — Add color op needs a dedicated additive shader per rendering backend (MonoGame/KNI/FNA, raylib, Skia) plus batching changes in Sprite/NineSlice, and no one has asked for it. Multiply/Alpha (the common case) already shipped in #4490. Reopen if a real use case comes up.

PR #4489, opened 2026-08-21
Support per-frame Alpha from .achx/.achj animation chains

### LINKED ISSUE #4488: Support per-frame Alpha from .achx/.achj animation chains
## Problem
`.achx`/`.achj` animation chains can carry a per-frame `Alpha` value (FlatRedBall2's `AnimationChain.Common` frame format), but Gum's `AnimationFrameSave` (`Gum/Graphics/Animation/Content/AnimationFrameSave.cs`) has no `Alpha` field at all, and the runtime `AnimationFrame`/`AnimationChainLogic` playback path (`RenderingLibrary/Graphics/Animation/`) has no code to apply it. An `.achx`/`.achj` file with per-frame alpha silently loses that data on load in the Gum tool.

This overlaps #4477 (per-frame Red/Green/Blue + ColorOperation), which lists Alpha among the missing fields, but Alpha is scoped independently there: unlike Multiply/Add color (which needs a per-backend pixel shader), Alpha only needs to scale the existing draw alpha — no shader work required. Alpha can be added without waiting on the rest of #4477.

## Suggested improvement
Add a nullable `Alpha` (0-255) field to `AnimationFrameSave`, parsed/written for both `.achx` and `.achj`. Apply it during Sprite/NineSlice frame playback by scaling the render alpha, same as FRB2's runtime application.

## Source
- `Gum/Graphics/Animation/Content/AnimationFrameSave.cs` — no `Alpha` field; add alongside the existing `FlipHorizontal`/`FlipVertical`/`FlipDiagonal` fields.
- `RenderingLibrary/Graphics/Animation/AnimationFrame.cs` — runtime frame type, also has no `Alpha` field.
- `RenderingLibrary/Graphics/Sprite.cs` (`ApplyAnimationFrame`) / `RenderingLibrary/Graphics/NineSlice.cs` (`ApplyAnimationFrame`) — where the frame's flip flags are currently applied on frame change; alpha application would go here.

## Reach
Any `.achx`/`.achj` file authored with per-frame alpha in the FlatRedBall Animation Editor loses that data when loaded into Gum.

vchelaru:
## Summary
- Adds a nullable `Alpha` (0-255) to `AnimationFrameSave`, parsed/written for both `.achx` (XML) and `.achj` (JSON).
- Threads `Alpha` through `AnimationFrame` into `Sprite`/`NineSlice` playback (`ApplyAnimationFrame`) — scales the existing `Alpha` property when the frame authors a value, left unchanged when the frame's `Alpha` is null (so authors can set it only on frames that need it, e.g. a flash effect).
- Scoped independently from #4477 (Red/Green/Blue + ColorOperation) since Alpha needs no per-backend shader work — just a draw-alpha scale.
- Also updates `MonoGameGumInCode`'s `SpriteScreen` sample (`AnimatedFrame1.achx`) to author per-frame Alpha, for manual verification.

Closes #4488

## Test plan
- [x] `dotnet test MonoGameGum.Tests/MonoGameGum.Tests.csproj` — 2192 passed
- [x] `KniGum`/`RaylibGum`/`SkiaGum` build clean, 0 new warnings
- [x] FRB canary (`GumCore.DesktopGlNet6.csproj`) — 0 errors
- [ ] Manual: MonoGameGumInCode's SpriteScreen, "AnimationChain-driven sprite" row — bear should flash dim (alpha 60) briefly each loop cycle

### DIFF
--- a/Gum/Graphics/Animation/Content/AnimationChainListSave.cs
+++ b/Gum/Graphics/Animation/Content/AnimationChainListSave.cs
@@ -202,7 +202,7 @@ namespace Gum.Content.AnimationChain
         // Property names match FlatRedBall2's AnimationChain.Common .achj writer (camelCase) so a
         // file authored by the FlatRedBall Animation Editor loads into Gum without a conversion
         // step. Fields FRB2 added that Gum's AnimationFrameSave doesn't model yet —
-        // Red/Green/Blue/Alpha, ColorOperation, per-frame shapes (#4477, #4479) — are simply
+        // Red/Green/Blue, ColorOperation, per-frame shapes (#4477, #4479) — are simply
         // not read; JsonObject's indexer returns null for a missing key, so those files still load,
         // just without that data.
         private static AnimationChainListSave ParseJson(JsonObject root)
@@ -256,6 +256,7 @@ namespace Gum.Content.AnimationChain
                 FlipDiagonal = BoolProperty(frameObj, "flipDiagonal"),
                 RelativeX = FloatProperty(frameObj, "relativeX"),
                 RelativeY = FloatProperty(frameObj, "relativeY"),
+                Alpha = IntProperty(frameObj, "alpha"),
             };
         }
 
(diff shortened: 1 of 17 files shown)


vchelaru:
Good catch from review: the original commit only wired Alpha into MonoGame-family's Sprite/NineSlice. Raylib and Skia have their own separate renderable classes with their own ApplyAnimationFrame — pushed a follow-up commit wiring Alpha into those too (both already had an Alpha property, just not read from the frame). SilkNetGum gets it for free since it references SkiaGum's renderables. Sokol intentionally skipped (not held to backend parity).

PR #4574, opened 2026-09-02
Add GumBatchDrawMode.Deferred to batch multiple Draw calls in one Begin/End cycle

### LINKED ISSUE #4573: GumBatch: opt-in Deferred draw mode to batch across Draw() calls within one Begin/End cycle
## Problem

`GumBatch`/`Renderer.Begin`/`Draw`/`End` (the immediate-mode entry point, `RenderingLibrary/Graphics/Renderer.cs:722-782`) never batches across multiple `Draw()` calls in the same `Begin`/`End` cycle, even though the layered path (`Renderer.RenderLayer`) can. Each top-level `Draw(IRenderableIpso)` call goes straight into the "legacy recursive walk" (`RenderingLibrary/Graphics/Renderer.cs:1118`, private `Draw(SystemManagers, Layer, IRenderableIpso)`), which submits that renderable's whole subtree to the GPU immediately and recursively. It never goes through the shared build-then-submit pipeline (`SiblingOrdering.BuildDrawList` + `Submit`) that `RenderLayer` uses for a `Layer`'s renderables, so `Renderer.SiblingOrdering = BatchKeyGroupedOrderer.Instance` has no effect across separate `Draw()` calls: it can only reorder within one already-submitted renderable's own subtree, not across siblings drawn in separate calls.

This matters for a host drawing many small entities through `GumBatch` (e.g. FRB2 drawing one playing-card entity per `Draw()` call): draw order across entities is fine for Z-sorting today, but nothing can batch their draw calls together even when they share a texture/`BatchKey`, because each `Draw()` call is its own immediate, unbatched submission.

## Suggested improvement

Add an opt-in `Deferred` mode, additive and off by default (mirrors MonoGame's own `SpriteSortMode.Immediate`/`Deferred` distinction — the reporter's framing):

- New enum with `Immediate` (default, today's exact behavior — no change) and `Deferred`.
- `GumBatch.Begin(Matrix? spriteBatchMatrix = null, mode = Immediate)` → `Renderer.Begin(spriteBatchMatrix, mode)`.
- In `Deferred` mode, `Draw()` still runs `InvokePreRenderRecursively` immediately (per-renderable), but appends the renderable to a new scratch list on `Renderer` instead of running the legacy recursive walk.
- `End()`, only when the cycle was `Deferred`: stable-sort that scratch list by `Z` (same algorithm as `Layer.SortRenderables`, `RenderingLibrary/Graphics/Layer.cs:98` — pull it into a small shared helper so both call sites use it), then run it through the existing subtree-entry-point overload `IRenderableOrderer.BuildDrawList(IList<IRenderableIpso> roots, ...)` (`RenderingLibrary/Graphics/IRenderableOrderer.cs:133`, already used by the render-target bake for exactly this "arbitrary list of top-level renderables, not a `Layer`" case) + `Submit(...)`, then clear the scratch list. `Immediate` mode is untouched.
- Do **not** reuse `_layers[0]`/`MainLayer` as the accumulation buffer — it's the same `Layer` the layered `Draw(SystemManagers)` path uses, so a host mixing both paths would corrupt it. Needs its own scratch list.
- Docs gotcha to carry over: `docs/code/rendering/gumbatch.md`'s "Mixing with SpriteBatch" section documents interleaving raw `gumBatch.SpriteBatch.Draw(...)` calls with `gumBatch.Draw(...)` calls, sharing submission order. That guarantee only holds in `Immediate` mode; `Deferred` mode needs the same caveat MonoGame gives for its own `SpriteSortMode.Deferred` (Gum's own draws no longer submit in call order relative to interleaved raw draws).

## Source

- `RenderingLibrary/Graphics/Renderer.cs:722` (`Begin`), `:735` (`Draw(IRenderableIpso)`), `:777` (`End`), `:1118` (legacy recursive walk comment + the private `Draw(SystemManagers, Layer, IRenderableIpso)` it calls).
- `RenderingLibrary/Graphics/Renderer.cs` `RenderLayer` (~line 649 area) — the existing build-then-submit sequence (`layer.SortRenderables()` → `SiblingOrdering.BuildDrawList(layer, ...)` → `Submit(...)`) this feature mirrors for the immediate-mode path.
- `RenderingLibrary/Graphics/IRenderableOrderer.cs:133` — the `roots`-based `BuildDrawList` overload to reuse.
- `RenderingLibrary/Graphics/Layer.cs:98` — `SortRenderables`, the stable-Z-sort to extract into a shared helper.
- `RenderingLibrary/Graphics/Renderer.cs:1653`/`:1663` (`GumBatch.Draw`/`End`) and `docs/code/rendering/gumbatch.md` — the public API surface and doc to update.

## Reach

Any `GumBatch`/immediate-mode consumer drawing many small entities per frame through separate `Draw()` calls (FRB2's per-entity draw pattern is the motivating case). Opt-in and additive — no existing behavior changes, so zero reach/risk for consumers who don't touch the new mode.

vchelaru:
## Summary

`Renderer.Begin`/`Draw`/`End` (what `GumBatch` wraps) submits each top-level `Draw()` call immediately and recursively, so nothing can batch across separate `Draw()` calls even when they share a texture/`BatchKey` - unlike the layered path, which sorts and builds one flat `DrawCommand` list per `Layer` before submitting. This matters for a host drawing many small entities through `GumBatch` (e.g. FRB2 drawing one playing-card entity per `Draw()` call): draws can't batch no matter what `SiblingOrdering` is set, since each call is its own isolated submission.

Adds an opt-in `Renderer.GumBatchDrawMode` (mirrors MonoGame's own `SpriteSortMode.Immediate`/`Deferred`), passed to `Begin(mode:)`, defaulting to `Immediate` (today's exact behavior, unchanged). In `Deferred` mode, `Draw()` accumulates renderables into a scratch list instead of submitting; `End` stable-sorts that list by `Z` (`Layer.SortByZ`, extracted out of `Layer.SortRenderables` so both share the algorithm) and runs it through the active `Renderer.SiblingOrdering` (e.g. `BatchKeyGroupedOrderer`) + `Submit` once, so separate `Draw()` calls can batch together.

Purely additive - no existing behavior changes for callers that don't pass `mode`.

Closes #4573

## Test plan

- [x] `LayerSortByZTests`: pins `Layer.SortByZ` correctness and stability (extraction didn't change behavior).
- [x] `GumBatchDeferredDrawModeTests`: `Immediate` (default) still one draw call per sprite across alternating textures; `Deferred` collapses to one draw call per distinct texture under `BatchKeyGroupedOrderer`.
- [x] Full `MonoGameGum.Tests` (2237) and `MonoGameGum.IntegrationTests` (110) suites green.
- [x] `KniGum`, `RaylibGum`, `SkiaGum` build clean (0 new warnings/errors).
- [x] FRB1 canary (`GumCore.DesktopGlNet6.csproj`, net6.0 + net8.0): pass, matches baseline.
- Manual test: needed - this changes real draw batching/order. Testing directly against a real FRB2 game (Solitaire) rather than a synthetic sample.

Claude-Session: https://claude.ai/code/session_01Nmz8Kz9gJyiP64APB87czb

### DIFF
--- a/.claude/skills/gum-monogame-rendering/SKILL.md
+++ b/.claude/skills/gum-monogame-rendering/SKILL.md
@@ -116,6 +116,8 @@ Three behaviors worth internalizing:
 
 **Landmine:** `BatchOrchestrator` only flushes on a `BatchKey` change, so its granularity caps what it can detect — a coarse key (e.g. one shared across many texture sources) means real per-texture draw-call cost hides inside MonoGame's own SpriteBatch batching over whatever order `SiblingOrdering` produced. Per-texture grouping goes through the separate `BatchSortKey` member instead, read only by `BatchKeyGroupedOrderer` — `BatchOrchestrator` never sees it, so it carries no flush cost under the default orderer.
 
+`BatchKeyGroupedOrderer.Instance` exposes `MergeBlockedByOverlapCount`/`NoCandidateInWindowBreakCount` (issue #4575), reset every `BuildDrawList` call — read them right after a draw to tell an overlap-forced batch break from genuine content alternation, instead of guessing from `GetDrawStateSummary` alone.
+
 ## SpriteBatchStack: Push / Pop / Replace
 
 `SpriteBatchStack` wraps a single `SpriteBatch` instance with a stack of `BeginParameters`:
--- a/.claude/skills/gum-monogame-rendering/SKILL.md
+++ b/.claude/skills/gum-monogame-rendering/SKILL.md
@@ -28,6 +28,8 @@ Key contract difference: each `Renderer.Draw` is one top-level renderable. If a 
 
 `Renderer.End` historically was asymmetric with `RenderLayer`'s end-of-walk: it called `EndSpriteBatch` but **did not** flush the pending custom batch, so draws leaked across cycles. Fixed: `Renderer.End` now calls `_batchOrchestrator.FlushAndReset(...)` before `EndSpriteBatch`. If you change the End logic, preserve that ordering.
 
+`Renderer.GumBatchDrawMode.Deferred` (optional `Begin(mode:)` param, issue #4573) accumulates `Draw()` calls into a scratch list instead of submitting immediately. `End` stable-sorts it by `Z` (`Layer.SortByZ`, extracted out of `Layer.SortRenderables`) and runs it through `SiblingOrdering.BuildDrawList`/`Submit` once, so separate `Draw()` calls can batch together (e.g. under `BatchKeyGroupedOrderer`). `Immediate` (the default) is unchanged.
+
 ## PreRender Walk: Layered Path Has Two Phases, GumBatch Path Has One
 
 The layered path runs a recursive `PreRender` pass on `layer.Renderables` **before** `BeginSpriteBatch`. That pass does two jobs:
@@ -286,3 +288,4 @@ When changing batch logic:
 9. If you touch `Renderer.AdjustRenderStates` or the clip-change paths in `Renderer.Draw`, does a clip change (entry OR exit) flush the orchestrator (`_batchOrchestrator.FlushAndReset`) BEFORE `BeginSpriteBatch`? Empirical canary: the first item's shape background inside a `ScrollViewer` clips when scrolled past the top edge (not just the second item and beyond).
 10. Investigating draw-call/batching cost? Check `Renderer.SiblingOrdering`, not just `BatchOrchestrator` — they're separate layers (see "Draw Order Is a Separate, Pluggable Layer" above).
 11. Reset immediate-mode perf stats once per host frame, not on every `Begin`. Per-`Begin` reset wipes out an earlier camera pass's stats instead of accumulating them within the frame.
+12. Touching the `Deferred` flush? It owns its own scratch buffers (`_deferredImmediateModeRoots`/`_deferredImmediateModeCommands`), not `_scratchCommands`/`_bakeCommands` — those are already claimed by `RenderLayer`/the render-target bake.
--- a/GumCommon/Collections/GraphicalUiElementCollection.cs
+++ b/GumCommon/Collections/GraphicalUiElementCollection.cs
@@ -79,13 +79,18 @@ public class GraphicalUiElementCollection : ObservableCollectionNoReset<Graphica
                 case NotifyCollectionChangedAction.Add:
                     if (e.NewItems != null)
                     {
-                        int index = e.NewStartingIndex;
+                        // e.NewStartingIndex is a RAW inner-collection index, which may also hold
+                        // non-GraphicalUiElement items this wrapper doesn't mirror (see
+                        // ToInnerIndex) - translate to the logical position among mirrored items
+                        // rather than assuming they line up.
+                        int rawIndex = e.NewStartingIndex;
                         foreach (var item in e.NewItems)
                         {
                             if (item is GraphicalUiElement gue)
                             {
-                                base.InsertItem(index++, gue);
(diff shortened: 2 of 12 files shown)


vchelaru:
Findings from a review pass while building #4580 on top of this branch. #4580 fixes the one in its own blast radius (a free container swallowing the clip-exit forced transition); the rest are left here since they're in this PR's code.

**1. `Renderer.End()` clears `ForcedMatrix` before the deferred flush — user-visible.** `spriteRenderer.ForcedMatrix = null` runs at the top of `End()`, but the `Deferred` block submits below it. `Immediate` submitted during `Draw()` while the matrix was still set; `Deferred` re-begins the batch on every clip enter/exit and orchestrator flush, now with `ForcedMatrix == null`. So `GumBatch.Begin(someMatrix, mode: Deferred)` over a tree containing a `ClipsChildren` container renders the clipped subtree at the camera transform instead of the forced one. Move the `= null` below the deferred block. `GumBatchDeferredDrawModeTests` has no matrix or clip case.

**2. `_deferredImmediateModeCommands` is never cleared after `Submit`** — holds `DrawCommand` references to the last cycle's renderables until the next deferred `End()`. Related: if `Submit` throws, `_deferredImmediateModeRoots.Clear()` is skipped and stale roots re-submit next cycle. try/finally, or clear before submitting.

**3. Deferred sort ignores `SecondarySortOnY`.** `Layer.SortByZ(_deferredImmediateModeRoots)` takes the default `secondarySortOnY: false` but submits into `_layers[0]`. If a consumer sets `_layers[0].SecondarySortOnY = true`, `Immediate` and `Deferred` disagree on equal-Z ordering. Pass `_layers[0].SecondarySortOnY`.

**4. `docs/code/performance-and-optimization/batchkeygroupedorderer.md:94` contradicts the code.** It says both counters "reset at the start of every `BuildDrawList` call"; they actually reset in `ResetBreakTally`, called once per host frame (immediate) or per `Draw(SystemManagers)` (layered) — as `MergeBlockedByOverlapCount`'s own XML doc says. Following the doc's snippet on the immediate path reads values accumulated across every Begin/End cycle in the frame. The section also doesn't mention `HardBoundaryTransitionCount`.

**5. `GraphicalUiElementCollection` — unguarded `IndexOf`.** At the two `_innerCollection.IndexOf(base.Items[index])` sites, a `-1` (wrapper and inner drifted apart) now throws from `RemoveAt(-1)` / `_innerCollection[-1] = item` rather than silently misbehaving. Worth a guard given the mirroring is what's being repaired here.

**6. `GraphicalUiElementCollection.ClearItems` wipes non-mirrored inner items.** This PR establishes that `_innerCollection` can hold raw non-`GraphicalUiElement` renderables (a shape's auto-wired stroke). `ClearItems` clears the whole inner collection, so `Children.Clear()` destroys that stroke and the shape then renders without it. Pre-existing, but the same ownership-bug family as the fix here, and the `Move`/`Replace` handlers are already flagged as latent.

vchelaru:
The findings in my earlier comment are now filed, since this PR has merged and a comment thread isn't tracking:

- #4582 — `ClearItems` wipes non-mirrored raw renderables (bug)
- #4583 — `Deferred` sort ignores `SecondarySortOnY` (bug)
- #4584 — `Deferred` block leaves stale roots queued if `Submit` throws (bug)
- #4585 — unguarded `IndexOf` result in `RemoveItem`/`SetItem`
- #4586 — `BatchKeyGroupedOrderer` docs state the wrong counter reset cadence

All five are flagged **unverified** — they came from reading code, not from an observed failure, and each says to prove it with a failing test before fixing (or close it if the test passes).

The sixth finding, `ForcedMatrix` being cleared before the `Deferred` submit, was proven with a red test and fixed in #4581.

PR #4671, opened 2026-09-10
Avalonia migration: ADR-0017, plan, coverage matrix, and the phase 10 canvas spike

kaltinril:
## Summary

Commits the Gum tool to a full Avalonia cutover (Windows, macOS, Linux native; WPF and WinForms retired at the end) and lays out how to get there.

- **ADR-0017** supersedes the deferral in ADR-0003. The decoupling decision itself stands and is the foundation this builds on.
- **Direction/avalonia-migration/**: primary plan (`README.md`), an inventory of the decoupling work already landed on main between June and September (`foundation.md`), a sweep of every Windows-only dependency in the tool graph with its owning phase and the guard that catches a regression (`coverage-matrix.md`), and thirteen phase docs.
- Roadmap moves the cross-platform editor from Later to Now; the open question is resolved; `ui-decoupling-plan.md` points its Phase 5 at the new plan.
- Target framework for the head is `net10.0`; a mechanical bump of the tool graph is the prerequisite before phase 20.
- **spikes/AvaloniaCanvasSpike/**: the phase 10 proof of concept. Not in any solution. One shared source renders a real Gum project through an XNA-family `Game` stepped from the Avalonia UI thread, reads the render target back, and presents it in an Avalonia window with pointer mapping, hit-element overlay, resize, and zoom. Two heads: MonoGame DesktopGL and KNI SDL2.GL. Headless render and frame-timing tests per backend.

## Verified on Windows

Both heads build with zero new warnings. Both headless tests pass (device with no host window, render, readback, hit test). Both Avalonia heads run and show the sample screen. Frame timing after zeroing `InactiveSleepTime` (the hidden game window is never active, and the XNA default sleeps 20 ms per tick):

| Size | MonoGame DesktopGL | KNI SDL2.GL |
|---|---|---|
| 1024x720 | 1.65 ms | 1.26 ms |
| 3840x2160 | 10.3 ms | 7.3 ms |

Readback is about 90% of the frame on both. macOS and Linux runs are still open and are the ones that retire the phase 10 risk.

## Process

Everything after this lands on main by ordinary PR beside the WPF head; there is no migration branch. The stalled `avalonia-port` branch is left as-is; its analysis is salvaged here, its code is not.


kaltinril:
Closing for now: the Avalonia migration stays on its branch until the whole migration is ready. This work is already on avalonia-migration-work.

PR #4682, opened 2026-09-12
Fix #4681: render the WPF tool canvas at physical-pixel resolution

### LINKED ISSUE #4681: Figure out if Gum scales according to monitor zoom, and if so, we should fix or document
https://discord.com/channels/586997072373481494/1073643744580079687/1547984896171970600

and 

https://discord.com/channels/586997072373481494/1073643744580079687/1547991873640075284

vchelaru:
Fixes the "1 pixel = 2 pixels" double-scaling reported in #4681: on a display scaled to e.g. 200%, the wireframe/texture-region canvas rendered into a `WriteableBitmap` sized in WPF device-independent units, and WPF's own DPI compositing then stretched that low-res bitmap to fill the physical-pixel area - blurring the canvas.

## Fix
- `WpfGraphicsDeviceControl` now sizes the render target/backing bitmap in physical pixels (DIU * DPI scale) instead of raw `ActualWidth`/`ActualHeight`, and stamps the bitmap's DPI to match so WPF displays it without an extra compositor stretch.
- `WireframeControl`/`ImageRegionSelectionControl` multiply their camera's zoom by that same DPI scale only for the duration of their draw call (`CameraDpiCompensationExtensions.BeginDpiCompensatedRender`), so the same logical content fills the larger physical-pixel target instead of only a fraction of it. Zoom and `ClientWidth`/`ClientHeight` are restored right after drawing, so hit-testing (which runs in DIU, matching WPF's own mouse coordinates) is unaffected.

## Tests
- `CameraDpiCompensationExtensionsTests` (new, `MonoGameGum.Tests`) - pins the Zoom/ClientWidth/ClientHeight scaling and restore math.
- `WpfGraphicsDeviceControlTests` (new) - pins the DIU-to-physical-pixel conversion.
- `WriteableBitmapRenderSurfaceTests`/`WpfRenderSurfaceHostTests` - updated for the new `dpiScale` parameter, plus new cases for the bitmap's DPI metadata.
- Full `GumToolUnitTests` (1361 passed) and `MonoGameGum.Tests` (2291 passed) suites green.

FRB canary: pass (`GumCore.DesktopGlNet6`, 0 errors - the new `CameraDpiCompensationExtensions.cs` is plain C# with no NET7+ APIs, added to `GumCoreShared.projitems` alongside `Camera.cs`).

Manual test: needed - display scaling can't be exercised by a unit test. Opening the Gum tool with the OS display scale set to e.g. 150%/200% and viewing the wireframe canvas should now show crisp content instead of blurry/doubled pixels.

Closes #4681

https://claude.ai/code/session_01B4Gro2Q1x6xuKJhSWs6R8z


vchelaru:
Pushed a second commit: the render-surface fix alone did nothing, because the Gum tool process had no DPI-awareness manifest declaration, so Windows was DPI-Unaware-virtualizing the whole window (bitmap-stretching it, and lying to WPF about its DIU size) - `VisualTreeHelper.GetDpi()` was reading 1.0 regardless of actual display scale, so the first fix never activated. Added `Gum/app.manifest` declaring Per-Monitor-V2 (confirmed via the built exe: no `dpiAwareness` block before, present after). That's also the real cause of the wrong *framing* (canvas drawing larger than the viewport) from the screenshots - not just blur.

vchelaru:
Pushed a third commit: removed the camera-zoom DPI compensation I'd added in the first commit. It was wrong - the render target is already physical-pixel-sized with matching bitmap DPI, so 1 render-target pixel = 1 physical screen pixel with no extra scaling needed. Multiplying Zoom on top of that doubled every world unit's on-screen size (100x100 rendering as 200x200) and desynced hit-testing from what was drawn (2x-speed dragging) - exactly the two symptoms reported. Not yet manually reconfirmed against a running build - that's the next step.

vchelaru:
Found it: hit-testing (Cursor/SelectionManager/drag handlers) reads coordinates from `WpfInputHostAdapter`/`MouseExtensions`, both of which still returned raw WPF DIU values - never converted for display scale, even though the render target (and Camera.ClientWidth/Height) are now physical-pixel-sized. On a 200% display that's exactly a 2x mismatch, matching "still zoomed 2x" for dragging. Pushed a fix multiplying by the live DPI scale at both entry points.

Also answering your TDD question directly on the issue/PR.

PR #4712, opened 2026-09-15
Fix #4710: Add Forms converts copied theme content to a .gumj project's own format

### LINKED ISSUE #4710: Add Forms should convert copied theme content to a .gumj project's own format
## Problem

Add Forms (both from the New Project flow and the standalone Add Forms dialog) copies a theme's Components/Screens/Behaviors/Standards into the project as raw `.gucx`/`.gusx`/`.behx`/`.gutx` XML files (`FormsThemeImporter.SaveFilesToDestination`, `FormsFileService.GetSourceDestinations`). For a `.gumj` (JSON) project, these copied files are never the ones the project actually loads:

- Components/Screens/Behaviors get correctly re-saved to `.gucj`/`.gusj`/`.behj` right afterward, because `ImportLogic`'s `TryAutoSaveElement`/`TryAutoSaveBehavior` calls already resolve the save path from the project's own format (fixed for #4595). So the project ends up correct, but the raw XML copies are left behind as dead, unreferenced files.
- Standards (`.gutx`, Standard theme only) are the one real correctness gap: they're picked up purely by the final full-project reload re-scanning `Standards/` on disk, which only looks for `.gutj` on a JSON project - so a copied `.gutx` is silently ignored. Verified with `gumcli diff-standards` that the Standard theme's Standards content is currently identical to `StandardElementsManager`'s bare defaults, so this has no observable effect **today** - but it would silently break the moment the Standard theme's Standards ever diverge from the defaults, and there's no test pinning that it doesn't.

Found while implementing #4705 (default new projects to `.gumj`), which makes this the *default* New Project path (Include Forms Controls is checked by default).

## Suggested improvement

Make `FormsThemeImporter`/`FormsFileService` format-aware for the destination project, similar to the fix applied to `FormsTemplateCreator` in #4705: convert copied Components/Screens/Standards/Behaviors to the destination project's format at copy time (or immediately after), and don't leave the XML copies behind for a `.gumj` project.

## Source

- `Tools/Gum.Presentation/GumForms/Services/FormsThemeImporter.cs` (`SaveFilesToDestination`, `AddAllElementsToProject`, `GetIfShouldSave`) - all extension-check logic assumes XML.
- `Tools/Gum.Presentation/GumForms/Services/FormsFileService.cs` (`GetSourceDestinations`) - computes destination paths always with the theme's own XML extensions.
- `Tools/Gum.ProjectServices/FormsTemplateCreator.cs` - the sibling fix for the CLI's own Forms template, reusable as a pattern (extract as XML, convert via `ConvertProjectToJsonService`, delete the XML).

## Reach

Every New Project created with "Include Forms controls" checked (the default) or Add Forms run against an existing project, when the destination project is `.gumj`. This is the primary, recommended way to start a Gum project - not a legacy corner case.

vchelaru:
## Summary

`FormsFileService.GetSourceDestinations` now computes a JSON-format destination path for a theme's Components/Screens/Standards/Behaviors when the destination project is `.gumj`, and `FormsThemeImporter` converts the theme's XML content to that format at copy time instead of leaving dead `.gucx`/`.gusx`/`.behx` copies behind.

Standards were the one real correctness gap: a copied `.gutx` was never routed through `ImportLogic`, so the final project reload (which only ever resolves `.gutj` for a JSON project) silently ignored it, leaving a `.gumj` project stuck with default Standards instead of the theme's own.

Also fixes `GumFormsLogic.GetIfProjectHasForms()`, found while extending the design: it hardcoded `.gutx` to exclude Standards from its "has Forms already" check. For a `.gumj` project every Standard destination is now `.gutj`, which would have made every JSON project falsely report "already has Forms" and permanently hide the Add Forms menu item.

Closes #4710

## Test plan
- [x] TDD: failing tests written first for each behavior change (Standards conversion, Component conversion + ImportLogic routing, `GetIfShouldSave`'s Standards-extension recognition, `GetIfProjectHasForms`'s Standards-extension recognition), confirmed red against the unfixed code, then green.
- [x] `dotnet test Tests/Gum.Presentation.Tests` — 1987/1987 passed.
- [x] `dotnet build GumFull.sln` — 0 errors, no new warnings.
- [x] Manual: Add Forms against a `.gumj` project in the tool, confirm no orphan `.gucx`/`.gusx`/`.behx`/`.gutx` files and Standards apply correctly.


vchelaru:
Also fixed an unrelated crash hit while manually testing this PR: `NullReferenceException` in `SelectionManager.HighlightedIpso` when the Avalonia head's tree view fires a hover event before `SelectionManager.Initialize()` has run (two-stage init race, Avalonia-specific). See the second commit.

vchelaru:
Also bundled a Gum.slnx fix hit while manually testing: GumPreview lived under Samples/ and wasn't part of any solution, so `dotnet build Gum.slnx` never built it and Preview failed with "Preview executable not found" until built separately. Added it to Gum.slnx. Follow-up filed to relocate it out of Samples/ entirely: #4713.

PR #4719, opened 2026-09-15
Fix #4717: spin the Preview icon on click, raise the preview window on re-click

### LINKED ISSUE #4717: Preview button should spin its icon on click and bring the preview window to the foreground
## Problem

The Preview (▶) button gives no feedback that clicking it did anything: the icon is a static glyph bound straight to `PreviewCommand`, and launching the GumPreview process takes a moment. Separately, if Preview is already running and the user clicks ▶ again on a different screen/component, `PreviewLauncher.Launch()` only pushes the new selection to the running process (`PreviewLauncher.cs:58-62`) — it never brings the GumPreview window to the foreground, so if that window got buried behind the editor, the user has no idea the selection actually changed.

## Suggested improvement

- Briefly spin/animate the ▶ icon when clicked, so there's visible confirmation the click registered while the process spawns.
- When Preview is already running and the button is clicked again, bring the GumPreview window to the foreground (in addition to the existing `PushSelection` call), not just silently update its content.

Open question for whoever picks this up: bringing another process's window to the foreground isn't a single cross-platform API (Win32 `SetForegroundWindow`/`AllowSetForegroundWindow` vs. X11/Wayland/macOS equivalents). GumPreview already polls the selection file (`PreviewLauncher.PushSelection`/`_selectionFilePath`), so it may be simplest to have GumPreview activate its own window when it detects a new selection, rather than the head trying to force-foreground another process's window itself.

## Source

- `Tool/Gum.Avalonia/Plugins/EditorTab/AvaloniaEditorTabPlugin.cs:215-226` — the Preview button, static "▶" content, bound to `EditorViewModel.PreviewCommand`
- `Tool/EditorTabPlugin.Core/ViewModels/EditorViewModel.cs:360` — `Preview()` command method calling `_previewLauncher.Launch()`
- `Tool/EditorTabPlugin.Core/Services/PreviewLauncher.cs:58-62` — the already-running branch that pushes the selection but never activates the window

## Reach

Every use of Preview (issue #4697/#4703) once a project is open — this is the button's only entry point, and the already-running path is hit on every re-click after the first launch.

vchelaru:
Closes #4717

## Changes
- `AvaloniaEditorTabPlugin.cs`: the ▶ Preview button now spins its icon one turn on click (a `RotateTransform` animated over 500ms via `DispatcherTimer`), so the click reads as registered while the GumPreview process spawns.
- `GumPreview/Game1.cs`: raises its own window (`SDL_RaiseWindow` via P/Invoke against `Window.Handle`, the SDL window handle on DesktopGL) whenever it detects a fresh write to the selection file - covers every re-click of Preview while already running, not just a changed selection. This resolves the issue's own open question: having GumPreview foreground itself is simpler and genuinely cross-platform (SDL abstracts Windows/X11/Wayland/macOS) versus the tool trying to force another process's window to the foreground.

## Manual test
Needed - runtime/visual UI change with no VM state to assert on.
1. Open a saved project, select a screen/component, click ▶ Preview - icon should spin briefly and GumPreview should launch.
2. Move the GumPreview window behind the tool (or another window).
3. Click ▶ Preview again (same or a different screen/component) - GumPreview's window should come to the foreground.

FRB canary: not needed - no FRB1-shared source touched (Avalonia-only view + the GumPreview sample host).


vchelaru:
Follow-up fixes based on manual testing feedback:
- Icon-only spinner (dashed ring, explicitly centered) instead of rotating the whole button.
- Preview button now disables when nothing is selected, instead of silently erroring to the Output tab.
- Selection-file protocol now carries an explicit activate marker: an explicit re-click of Preview raises the GumPreview window, but browsing the tool's tree (which still live-updates the preview content via the same file) no longer steals focus.

Manual test steps updated:
1. With nothing selected, ▶ should be disabled (greyed out).
2. Select a screen/component - ▶ enables, click it - icon spinner plays, GumPreview launches.
3. Move GumPreview behind the tool. Click a *different* screen/component in the tree - preview content updates but the window stays put (no focus steal).
4. Click ▶ again - GumPreview comes to the foreground.

PR #4844, opened 2026-09-18
Fix #4837: Ctrl+Shift-click a top-level Component/Screen to add it as a child of the current selection

### LINKED ISSUE #4837: Ctrl+Shift-clicking a top-level component should add it as a child instance of the current selection
## Problem

Holding Ctrl+Shift while clicking a top-level component in the tree currently does nothing special — it falls through to a plain single-select, same as an unmodified click (`TreeNodeClickDispatchLogic.GetReaction` only special-cases `Control` and `Shift` individually, not the combined flag). The user wants this combo to instead add an instance of the clicked component as a child of whatever is currently selected (a screen, component, or instance) — a keyboard shortcut for what dragging that component onto the selection already does.

## Suggested improvement

When Ctrl+Shift-click lands on a top-level component node and something else (screen/component/instance) is already selected, add a new instance of the clicked component as a child of the current selection, subject to the same rules drag-and-drop already enforces: no self/descendant (circular) drops, and `ICircularReferenceManager.CanTypeBeAddedToElement` must allow it. Invalid combinations should just no-op (or show the same rejection the drag path shows), not throw or partially add.

Open question for whoever implements this: does "component" here mean components only, or should top-level screens behave the same way? The report only mentions components.

## Source

- `Tool/TreeViewPlugin.Core/TreeSelection/TreeNodeClickDispatchLogic.cs:42-67` — click-reaction dispatch; `TreeModifierKeys` is `[Flags]` (`TreeInputTypes.cs`) so Control+Shift is already a distinct combined value here, just unhandled.
- `Tools/Gum.Presentation/Managers/DragDropManager.cs:374` — the existing drag-drop path's circular-reference gate (`_circularReferenceManager.CanTypeBeAddedToElement`) that the new path should reuse rather than duplicate.
- `Tool/TreeViewPlugin.Core/ElementTreeViewManager.cs:476` (`AddStandardInstanceToCurrentElement`) — an existing non-drag "click to add instance to current selection" precedent (used by the Standards palette chips) worth reusing/mirroring.

## Reach

Tree view only (component/screen/instance list), not the canvas. Any user who knows the shortcut can reach it on every top-level component; it's not a hand-authored-file corner case.

vchelaru:
## Summary
- Ctrl+Shift-click on a top-level Component or Screen node in the tree adds an instance of it as a child of whatever is currently selected (screen/component/instance), reusing the exact same drag-drop path (`DragDropManager.HandleDroppedElementOnTreeNode`, widened from Standards-only to any `ElementSave`) that a real drag onto the target already uses - so circular-reference checks, "Screens can't be instantiated," wrong-state-selected, etc. are enforced identically, with the same rejection dialog.
- Splits the Standards palette chip's Ctrl+Shift-click from its pre-existing Ctrl+click (which always added to the open element's root) per the issue's follow-up comment - Ctrl+Shift now adds as a child of the current tree selection instead.
- Screens are handled uniformly with Components (no special-casing needed): the reused drag-drop path already rejects a Screen target with "Screens can't be dropped into other Screens or Components."

Closes #4837

## Test plan
- [x] New unit tests: `TreeNodeAddAsChildClickLogicTests`, `TreeSelectionModelAddAsChildTests`, `DragDropManagerHandleDroppedElementOnTreeNodeTests` (Tests/Gum.Presentation.Tests) - all green, confirmed red-first for the new `TreeSelectionModel` branch.
- [x] Existing `DragDropManagerTests` (GumToolUnitTests, renamed calls) still pass; full `Gum.Presentation.Tests` (2094) and `Gum.Avalonia.Tests` (171) suites green; `Gum.Wpf.sln` and `Gum.slnx` both build with zero new warnings.
- [ ] Manual: Ctrl+Shift-click a top-level Component onto a selected Screen/Component/instance and confirm it's added as a child; confirm a Screen target shows the existing rejection dialog; confirm the Standards palette's Ctrl+Shift+click on a chip adds as a child of the selection while plain Ctrl+click still adds to the open element's root.


vchelaru:
Follow-up pushed: plain Ctrl+click and Ctrl+Shift+click on a Standards chip now go through the identical code path (add as a child of the current selection). Previously they were split per the issue's earlier follow-up comment, but that split itself turned out to be the wrong call - Victor flagged it as inconsistent during manual testing. Ctrl+Shift+click on a chip is now redundant with plain Ctrl+click; the explicit root-level add is still available via the chip's right-click "Add to <element>" menu item.

PR #4883, opened 2026-09-19
Ctrl-click/Ctrl+Shift-click add-instance gestures no longer scroll the tree view

### LINKED ISSUE #4882: Adding an instance via Ctrl+Click/Ctrl+Shift+Click should not scroll the tree view into view
**Problem**
Ctrl+Shift+Click on an element node (and Ctrl+Click/Ctrl+Shift+Click on a Standards chip) adds a new instance and selects it — good — but the tree view also scrolls to bring the new instance into view. That scroll is normally desirable, but for this specific gesture it interrupts users adding several instances in a row, since each add jerks the tree view's scroll position.

**Suggested improvement**
Suppress the scroll-into-view specifically for instances created by these add gestures, while keeping it for other selection changes (click, keyboard nav, etc.), so users can keep adding instances one after another without the view jumping.

**Source**
Selection changes (including the one driven by adding an instance) flow into `ElementTreeViewManager.Select(GumTreeNode?)` (`Tool/TreeViewPlugin.Core/ElementTreeViewManager.cs:1283-1307`), which unconditionally calls `View.EnsureVisible(treeNode)` at line 1299. The add-instance gestures themselves go through `IAddInstanceLogic.AddInstance`/`AddInstanceAtDestination` (`Tools/Gum.Presentation/Logic/AddInstanceLogic.cs:47-76`), which sets `_selectedState.SelectedElement`/the new selection but has no direct tree-view reference. The fix likely needs a way to flag "this selection change came from an add-instance gesture" so `Select` can skip `EnsureVisible` just for it.

**Reach**
This is the normal, tool-recommended way to add instances repeatedly (chip Ctrl+click, Ctrl+Shift+click on an element node) — not a rare or legacy path.

vchelaru:
Closes #4882.

Ctrl-click on a Standards chip and Ctrl+Shift-click on an element node add an instance and select it — the selection unconditionally scrolled the tree into view, interrupting users adding several instances in a row.

`ElementTreeViewManager.Select` now skips `EnsureVisible` while an internal `SuppressNextEnsureVisible` flag is set; the two gestures bracket their `IAddInstanceLogic` call with it. Drag, the right-click Add menu, and the Add Instance dialog are left alone — each is a single deliberate placement, not a rapid-fire sequence the scroll interrupts.


vchelaru:
Follow-up: fixed a related virtualization gap the scroll-suppression fix surfaced — inserting a tree row above the current scroll viewport (e.g. adding an instance to a container scrolled out of view) shifted every already-visible row down by one, since the scroll offset was never adjusted. `AvaloniaGumTreeView` now anchors the topmost visible row across a rebuild and restores the scroll offset to keep it in the same on-screen position. Covered by a new headless test (`InsertingARowAboveTheViewport_KeepsAlreadyVisibleRowsAtTheSameScreenPosition`), verified red/green against the fix.

